PR #2, opened 2020-02-14
Add write registers command

iburinoc:
Necessary for gdb to take control and run some commands to extract data

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -69,6 +69,12 @@ pub trait Target {
     /// register's value.
     fn read_registers(&mut self, push_reg: impl FnMut(&[u8]));
 
+    /// Write the target's registers.
+    ///
+    /// The registers are provided in the order specified in the target's
+    /// `target.xml`.
+    fn write_registers(&mut self, regs: impl FnMut() -> Option<Self::Usize>);
+
     /// Read the target's current PC
     fn read_pc(&mut self) -> Self::Usize;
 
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -69,6 +69,12 @@ pub trait Target {
     /// register's value.
     fn read_registers(&mut self, push_reg: impl FnMut(&[u8]));
 
+    /// Write the target's registers.
+    ///
+    /// The registers are provided in the order specified in the target's
+    /// `target.xml`.
+    fn write_registers(&mut self, regs: impl FnMut() -> Option<Self::Usize>);
```
daniel5151: `read_registers` makes no assumptions about the size of registers (it's callback accepts a slice of u8s, which can be any size). This is deliberate, as some platforms have variable-sized registers (e.g: x86's floating point and vector extensions)

`write_registers` needs to provide the target with the raw `&[u8]` that was parsed from the packet (or potentially an iterator u8, if you want to be fancy), leaving the interpretation up to the individual targets.

src/stub.rs
```diff
@@ -104,6 +104,21 @@ impl<T: Target, C: Connection> GdbStub<T, C> {
                 });
                 err?;
             }
+            Command::G(cmd) => {
```
daniel5151: this implementation will have to be modified based off the comment above

daniel5151:
👏

PR #6, opened 2020-03-28
Allow target to trigger breakpoints itself

sapir:


src/lib.rs
```diff
@@ -374,6 +374,8 @@ pub enum TargetState {
     Running,
     /// Halted
     Halted,
+    /// Running but hit a breakpoint
```
daniel5151:
```suggestion
    /// Running, but hit a breakpoint
```

src/stub.rs
```diff
@@ -359,7 +359,8 @@ impl<T: Target, C: Connection> GdbStub<T, C> {
 
                     // TODO: defer implementation of hardware and software breakpoints to target
                     let target_pc = target.read_pc();
-                    if self.swbreak.contains(&target_pc) {
+                    if target_state == TargetState::Breakpoint || self.swbreak.contains(&target_pc)
+                    {
```
daniel5151:
```suggestion
                    if target_state == TargetState::Breakpoint || self.swbreak.contains(&target_pc) {
```

daniel5151:
If you take a look at the GDB Remote Serial Protocol docs about [Stop Reply Packets](https://sourceware.org/gdb/current/onlinedocs/gdb/Stop-Reply-Packets.html#Stop-Reply-Packets), you'll find that there are _many_ different reasons a target might suddenly stop. Notably, there is a difference between hardware and software breakpoints. As such, I don't think `Breakpoint` is specific enough to describe this action.

Could you change the variant's name and documentation to reflect that it's triggering a software breakpoint?

sapir:
םל

sapir:
ok

PR #7, opened 2020-04-06
Fix infinite loop and stack overflow on call to <Error as Debug>::fmt

sapir:


daniel5151:
Nice catch! I'll merge and push a hotfix.

daniel5151:
pushed to crates.io as part of 0.1.2

l33ds:
Thank you @sapir , you rock

PR #11, opened 2020-08-18
Add x86_64 support

jamcleod:
One potential concern is the naming, which I'm fine with changing. Currently I have:

**module:** `x86`
**Arch struct:** `X86_64`
**Registers struct:** Amd64CoreRegs

### DIFF
--- a/src/arch/x86/reg/amd64_core.rs
+++ b/src/arch/x86/reg/amd64_core.rs
@@ -1,0 +1,155 @@
+use crate::arch::Registers;
+use core::convert::TryInto;
+
+/// 64-bit x86 core registers.
+///
+/// Source: https://github.com/bminor/binutils-gdb/blob/master/gdb/features/arm/arm-core.xml
+#[derive(Default)]
+pub struct Amd64CoreRegs {
+    // RAX, RBX, RCX, RDX, RSI, RDI, RBP, RSP, r8-r15
+    regs: [u64; 16],
+    // Status register
+    eflags: u32,
+    // Instruction pointer
+    rip: u64,
+    // Segment registers: CS, SS, DS, ES, FS, GS
+    segments: [u32; 6],
+    // FPU registers: ST0 through ST7
+    st_regs: [F80; 8],
+    // SIMD Registers: XMM0 through XMM15
+    xmm_regs: [u128; 0x10],
+    // SSE Status/Control Register
+    mxcsr: u32,
+}
+
+type F80 = [u8; 10];
+
+impl Registers for Amd64CoreRegs {
+    fn gdb_serialize(&self, mut write_byte: impl FnMut(Option<u8>)) {
+        macro_rules! write_bytes {
+            ($bytes:expr) => {
+                for b in $bytes {
+                    write_byte(Some(*b))
+                }
+            };
+        }
+
+        for reg in &self.regs {
+            write_bytes!(&reg.to_le_bytes());
+        }
+
+        // rip
+        write_bytes!(&self.rip.to_le_bytes());
+
+        // eflags
+        write_bytes!(&self.eflags.to_le_bytes());
+
+        // cs, ss, ds, es, fs, gs
+        for seg in &self.segments {
+            write_bytes!(&seg.to_le_bytes());
+        }
+
+        // st0 to st7
+        for st_reg in &self.st_regs {
+            write_bytes!(st_reg);
+        }
+
+        // fctrl, fstat, ftag, fiseg, fioff, foseg, fooff, fop
+        for _ in 0..8 {
+            write_bytes!(&[0; 4]);
+        }
+
+        // xmm0 to xmm15
+        for xmm_reg in &self.xmm_regs {
+            write_bytes!(&xmm_reg.to_le_bytes());
+        }
+
+        // mxcsr
+        write_bytes!(&self.mxcsr.to_le_bytes());
+
+        // padding?
+        write_bytes!(&[0x10; 0x18]);
+    }
+
+    fn gdb_deserialize(&mut self, bytes: &[u8]) -> Result<(), ()> {
+        let mut regs = bytes[0..0x80]
+            .chunks_exact(8)
+            .map(|x| u64::from_le_bytes(x.try_into().unwrap()));
+
+        for reg in self.regs.iter_mut() {
+            *reg = regs.next().unwrap();
+        }
+
+        self.rip = u64::from_le_bytes(bytes[0x80..0x88].try_into().unwrap());
+        self.eflags = u32::from_le_bytes(bytes[0x88..0x8C].try_into().unwrap());
+
+        let mut segments = bytes[0x8C..0xA4]
+            .chunks_exact(4)
+            .map(|x| u32::from_le_bytes(x.try_into().unwrap()));
+
+        for seg in self.segments.iter_mut() {
+            *seg = segments.next().unwrap();
+        }
+
+        let mut regs = bytes[0xA4..0xF4]
+            .chunks_exact(10)
+            .map(|x| x.try_into().unwrap());
+
+        for reg in self.st_regs.iter_mut() {
+            *reg = regs.next().unwrap();
+        }
+
+        let mut regs = bytes[0x114..0x214]
+            .chunks_exact(0x10)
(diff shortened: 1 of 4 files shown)


src/arch/x86/reg/amd64_core.rs
```diff
@@ -0,0 +1,155 @@
...
+    // Status register
+    eflags: u32,
+    // Instruction pointer
+    rip: u64,
+    // Segment registers: CS, SS, DS, ES, FS, GS
+    segments: [u32; 6],
+    // FPU registers: ST0 through ST7
+    st_regs: [F80; 8],
+    // SIMD Registers: XMM0 through XMM15
+    xmm_regs: [u128; 0x10],
```
daniel5151, line 20: hmm, `xmm` and `mxcsr` aren't technically part of `i386/64bit-core.xml`, and seem to be defined in `i386/64bit-sse.xml`.

That said, I kinda fudged things myself when implementing ARM, and included the FPU registers (defined in `arm-fpa.xml`) in the "core" registers, so I guess we can just fudge it for x86 as well...

src/arch/x86/reg/amd64_core.rs
```diff
@@ -0,0 +1,155 @@
...
+
+        // xmm0 to xmm15
+        for xmm_reg in &self.xmm_regs {
+            write_bytes!(&xmm_reg.to_le_bytes());
+        }
+
+        // mxcsr
+        write_bytes!(&self.mxcsr.to_le_bytes());
+
+        // padding?
```
daniel5151, line 70: uhhhh, could you dig into what these bytes actually represent? 😬

src/arch/x86/reg/amd64_core.rs
```diff
@@ -0,0 +1,155 @@
...
+        }
+
+        // st0 to st7
+        for st_reg in &self.st_regs {
+            write_bytes!(st_reg);
+        }
+
+        // fctrl, fstat, ftag, fiseg, fioff, foseg, fooff, fop
+        for _ in 0..8 {
+            write_bytes!(&[0; 4]);
```
daniel5151, line 59: I'm not too familiar with x86 innards, are these registers relatively unused nowadays? If so, I don't mind leaving them stubbed out, though if it's not too much trouble, [de]serializing them would be nice.

src/arch/x86/reg/amd64_core.rs
```diff
@@ -0,0 +1,155 @@
...
+    // RAX, RBX, RCX, RDX, RSI, RDI, RBP, RSP, r8-r15
+    regs: [u64; 16],
+    // Status register
+    eflags: u32,
+    // Instruction pointer
+    rip: u64,
+    // Segment registers: CS, SS, DS, ES, FS, GS
+    segments: [u32; 6],
+    // FPU registers: ST0 through ST7
+    st_regs: [F80; 8],
```
daniel5151, line 18: Adding `_regs` is unnecessary

src/arch/x86/reg/amd64_core.rs
```diff
@@ -0,0 +1,155 @@
...
+        write_bytes!(&self.mxcsr.to_le_bytes());
+
+        // padding?
+        write_bytes!(&[0x10; 0x18]);
+    }
+
+    fn gdb_deserialize(&mut self, bytes: &[u8]) -> Result<(), ()> {
+        let mut regs = bytes[0..0x80]
+            .chunks_exact(8)
+            .map(|x| u64::from_le_bytes(x.try_into().unwrap()));
```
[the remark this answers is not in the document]
jamcleod: In this situation, the unwrap panic is unreachable without a serious error on the part of the programmer, as both the indexing and `chunks_exact` would cause issues before that unwrap is reachable in situations which would cause a panic.

I have, however, done my best to remove all the unwraps, however I wasn't able to get the last one (added alongside all the nasty FPU internal registers) without making the code significantly more verbose.

I honestly assumed this style of unwrapping was fine due to it being used [elsewhere in the codebase](https://github.com/daniel5151/gdbstub/blob/master/src/arch/arm/reg/arm_core.rs#L54)

src/arch/x86/reg/amd64_core.rs
jamcleod: I did my best (and tried again)... all registers GDB displays with `info all-registers` are accounted for, and no combination of poking at various gdb features, looking at GDB source, and filling that section with various forms of data resulted in me figure out what those bytes are. All I know is that if I include more or less bytes, GDB gets mad. I would love to find out myself, but GDB is not my strongest suit so maybe I'm missing something...

(also, admittedly, I ended up just throwing data at GDB and reverse engineering how it decodes it because I found the x86 target description very hard to correlate into an actual specification of layout)

src/arch/x86/reg/amd64_core.rs
jamcleod: Ok so I took a second stab at those because GDB has very non-standard names for them on top of them basically being internal registers for the x87 FPU (which virtually nobody really uses anyway) it was rather difficult, but after re-reading all of the FPU section of the intel manual on it, I managed to figure out what the heck was up with the GDB naming scheme. Should be sane now.

jamcleod:
Also thanks for the quick and thorough review!

src/arch/x86/reg/amd64_core.rs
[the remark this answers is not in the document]
daniel5151: Ah, you know what, I'm a dummy. 
You're totally right, and in-fact, this is the one unwrap case that totally makes sense!

What I meant to convey was unwrapping in the _rest_ of the function wasn't necessary. That's on me 😅

Let's re-add that key unwrap, and make the rest of the code a lot less verbose, yeah? That aught to let us remove all those nasty `map_err(|_|())?` littered about.

src/arch/x86/reg/amd64_core.rs
daniel5151: All good, I feel your pain haha. 
GDB is quite the beast, and trying to figure out what's going on can be quite tricky at times.

I'm okay with leaving it in. Maybe add brief `// HACK:` (or `// XXX:`) comment noting your efforts into deciphering these bytes, and that while they must mean _something_, it doesn't seem to be something particularly important.

Additionally, instead of calling `write_bytes!(&[0; 0x18])`, could you replace it with a `(0..0x18).for_each(|_| write_byte(None))` instead? This will serialize as `xxxx...` bytes, indicating that these registers haven't been collected (as opposed to indicating that they've been collected, albeit with a zero value).

src/arch/x86/reg/amd64_core.rs
daniel5151: Ah, nice sleuthing!

src/arch/x86/reg/core64.rs
```diff
@@ -0,0 +1,148 @@
+use crate::arch::Registers;
+use crate::arch::x86::reg::F80;
+use core::convert::TryInto;
+
+/// 64-bit x86 core registers.
```
daniel5151, line 5:
```suggestion
/// 64-bit x86 core registers (+ SSE extensions).
```

src/arch/x86/reg/core64.rs
```diff
@@ -0,0 +1,148 @@
...
+///
+/// Source: https://github.com/bminor/binutils-gdb/blob/master/gdb/features/i386/64bit-core.xml
```
daniel5151, line 7: Additionally: https://github.com/bminor/binutils-gdb/blob/master/gdb/features/i386/64bit-sse.xml

src/arch/x86/reg/core64.rs
```diff
@@ -0,0 +1,148 @@
...
+    /// Status register
+    pub eflags: u32,
+    /// Instruction pointer
+    pub rip: u64,
+    /// Segment registers: CS, SS, DS, ES, FS, GS
+    pub segments: [u32; 6],
+    /// FPU registers: ST0 through ST7
+    pub st: [F80; 8],
+    /// Floating-point control register
+    pub fctrl: u32,
```
daniel5151, line 21: Thanks for looking into these floating-point registers!

Since they don't seem to be used very often nowadays, instead of cluttering up the main registers struct, lets extract these registers into a separate `struct X87FPUControlRegs` register. Feel free to tweak the naming if you like.

PR #19, opened 2020-08-26
Allow stop packets with arbitary signal codes

mchesser:
Currently `gdbstub` only allows SIGTRAP signals to be used in stop reply packets, however there are other signals that are useful to report (in particular I want to report SIGSEGV).

This PR partially issue report with _a possible_ way of allowing this. However, maybe you want to have explicit enum variants for all reasonable stop reasons, and this technically a breaking change, so may need to wait to the next non-patch release.

### DIFF
--- a/src/gdbstub_impl/mod.rs
+++ b/src/gdbstub_impl/mod.rs
@@ -567,6 +567,11 @@ impl<T: Target, C: Connection> GdbStubImpl<T, C> {
                 res.write_str("S05")?;
                 Ok(None)
             }
+            StopReason::Signal(code) => {
+                res.write_str("S")?;
+                res.write_num(code)?;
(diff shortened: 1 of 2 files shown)


daniel5151:
Yep, this seems reasonable to me. Thanks for the PR!

Technically, semver lets me make breaking changes before 1.x.y, but if it's not urget, I'd rather lump this in with some other changes I'm making for 0.3.

That said, I'll be marking `StopReason` as `#[non_exhaustive]` in future versions, which aught to allow merging these sorts of changes in a non-breaking manner.

daniel5151:
Also, please ignore the misc. branch thrashing that I just did. 
All latest development will happen on the master branch moving forward.

Issue #12, opened 2020-08-18
Support fine-grained control over Target Features


daniel5151:
Aside from specifying the core architecture of the target, the [Target Description XML](https://sourceware.org/gdb/current/onlinedocs/gdb/Target-Description-Format.html#Target-Description-Format) is also used to specify various _features_ of the architecture. As the docs say: "Features are currently used to describe available CPU registers and the types of their contents"

e.g: for x86, the core registers are defined under the `org.gnu.gdb.i386.core` feature, while the optional SSE registers are defined under the `org.gnu.gdb.i386.sse` feature.

Currently, the `Arch` trait isn't very well suited to describing features, requiring a separate implementation for each set of features. i.e: if an architecture has 3 different features, A, B, and C, each with their own registers, there would have to be 3! (i.e: 6) different register structs to describe each set of features!

I'd like to preserve the current "described at compile time" approach to the `Arch` trait, and I think with a little bit of experimentation, it shouldn't be too hard to find some sort of ergonomic trait-based solution to the problem. One thing that's for certain is that `target_features_xml` should be changed from returning a `Option<'static str>`, to possibly accepting a `FnMut(&str)` method which supports building up a XML file in chunks...

daniel5151:
Hmm, I wonder if a quick-and-dirty solution might be to create some sort of "arch builder" macro?

Instead of exposing types that implement `Arch` directly, each arch exposes a bunch of different "feature" structures, which encode a feature's associated registers + xml string. Consumers of the crate will then invoke the arch builder macro with the set of features they're interested in, creating a brand-new `Arch` impl suited specifically for their project.

A rough sketch of what the end-user API might look like:

```rust
arch_builder! {
    name: CustomX86,
    arch: x86_64,
    features: [
        core, sse, segments, avx
    ]
}
```

Not sure what the implementation might look like, but this would fix the immediate problem of how to easily compose different Arch traits.

Of course, this approach isn't _that_ great, since macro-based codegen is kinda hacky in my book. That said, I wanted to write this idea down before I forgot it, since it might be a good jumping off point in the future.

daniel5151:
So here's something interesting: https://sourceware.org/gdb/onlinedocs/gdb/Target-Description-Format.html#Inclusion

Turns out target description XMLs support using `<xi:include href="foo.xml"/>` to split the description into multiple files...

For example:

**target.xml**

```xml
<?xml version="1.0"?>
[...]

PR #21, opened 2020-08-30
Add (partial) support for RISC-V

thomashk0:
Hey,

This PR contains two commits:
- A first that implements the `Arch` trait for RISC-V 32/64 bit
- A second one that add support for the "Read register command" of the RSP protocol ('p' command), this is actually needed for debuging RISC-V programs. GDB is querying some specific CSR registers through this command (mstatus, mpriv, ...).

Currently I have a RISC-V softfp toolchain for testing. I will try to add support for floating-point RISC-V 32 and 64 as well soon.
Let me know if the commits need some rework.

Thanks for your awesome `gdbstub` crate, it works like a charm on my emulator !
Thomas.

### DIFF
--- a/src/arch/riscv/mod.rs
+++ b/src/arch/riscv/mod.rs
@@ -1,0 +1,33 @@
+//! Support for the [RISC-V](https://riscv.org/) architecture.
+//!
+//! *Note*: currently only supports integer version of the ISA
+
+use crate::arch::Arch;
+
+pub mod reg;
+
+/// Implements `Arch` for 32-bit RISC-V.
+#[derive(Eq, PartialEq)]
+pub struct Riscv32;
+
+/// Implements `Arch` for 64-bit RISC-V.
+#[derive(Eq, PartialEq)]
+pub struct Riscv64;
+
+impl Arch for Riscv32 {
+    type Usize = u32;
+    type Registers = reg::RiscvCoreRegs<u32>;
+
+    fn target_description_xml() -> Option<&'static str> {
+        Some(r#"<target version="1.0"><architecture>riscv</architecture></target>"#)
+    }
+}
+
+impl Arch for Riscv64 {
+    type Usize = u64;
+    type Registers = reg::RiscvCoreRegs<u64>;
+
+    fn target_description_xml() -> Option<&'static str> {
+        Some(r#"<target version="1.0"><architecture>riscv64</architecture></target>"#)
+    }
+}
--- a/src/arch/riscv/mod.rs
+++ b/src/arch/riscv/mod.rs
@@ -1,6 +1,6 @@
 //! Support for the [RISC-V](https://riscv.org/) architecture.
 //!
-//! *Note*: currently only supports integer version of the ISA
+//! *Note*: currently only supports integer versions of the ISA.
 
 use crate::arch::Arch;
 
--- a/src/arch/riscv/reg/riscv.rs
+++ b/src/arch/riscv/reg/riscv.rs
@@ -1,19 +1,18 @@
-//! RISC-V Register Definitions.
-//!
-//! Useful links:
-//!     - [GNU binutils-gdb XML Descriptions](https://github.com/bminor/binutils-gdb/blob/master/gdb/features/riscv)
-//!     - [riscv-tdep.h](https://github.com/bminor/binutils-gdb/blob/master/gdb/riscv-tdep.h)
 use crate::arch::Registers;
 use crate::internal::LeBytes;
 use num_traits::PrimInt;
 
-/// RISC-V Integer registers
+/// RISC-V Integer registers.
 ///
 /// The register width is set to `u32` or `u64` based on the `<U>` type.
+///
+/// Useful links:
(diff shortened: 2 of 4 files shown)


daniel5151:
Hey Thomas, thanks for the PR! 
Always glad to see another project successfully integrating `gdbstub`!

I'm don't think I'll get the chance to give the code a proper review today, but from a quick glance, the `Arch` implementation looks pretty good!

That said, could you please split the `read_register` functionality into it's own PR? The proposed API is leaking part of the underlying GDB protocol (namely, the `reg_number: usize` parameter), which is something `gdbstub` has tried to avoid. I've already got some ideas on how to tweak it, which I'll leave in a comment on the new PR once it's open.

thomashk0:
Thanks! 

I forced-pushed the `Arch` trait-only in this PR.
The new PR for discussing the read register API is here: #22 .

src/arch/riscv/reg/riscv.rs
```diff
@@ -0,0 +1,70 @@
...
+use crate::internal::LeBytes;
+use num_traits::PrimInt;
+
+/// RISC-V Integer registers
+///
+/// The register width is set to `u32` or `u64` based on the `<U>` type.
+#[derive(Default)]
+pub struct RiscvCoreRegs<U> {
+    /// General purpose registers (x0-x31)
+    pub r: [U; 32],
```
daniel5151, line 16: If the convention is to denote registers as `x0-x31`, then we might want to change this field from `pub r` to `pub x`.

src/arch/riscv/reg/riscv.rs
```diff
@@ -0,0 +1,70 @@
+//! RISC-V Register Definitions.
+//!
+//! Useful links:
+//!     - [GNU binutils-gdb XML Descriptions](https://github.com/bminor/binutils-gdb/blob/master/gdb/features/riscv)
```
daniel5151, line 4: Could you move these links down into the doc comment for `RiscvCoreRegs`? 
I'd like to keep things consistent with the other implementations.

Also, since the module isn't marked `pub`, this doc-comment won't actually correspond to any generated docs, so it can just be removed.

src/arch/riscv/mod.rs
```diff
@@ -0,0 +1,33 @@
+//! Support for the [RISC-V](https://riscv.org/) architecture.
+//!
+//! *Note*: currently only supports integer version of the ISA
```
daniel5151, line 3: nit: missing period

src/arch/riscv/reg/riscv.rs
```diff
@@ -0,0 +1,70 @@
...
+//!     - [riscv-tdep.h](https://github.com/bminor/binutils-gdb/blob/master/gdb/riscv-tdep.h)
+use crate::arch::Registers;
+use crate::internal::LeBytes;
+use num_traits::PrimInt;
+
+/// RISC-V Integer registers
```
daniel5151, line 10: nit: missing period

src/arch/riscv/reg/riscv.rs
```diff
@@ -0,0 +1,70 @@
...
+        // Read GPRs
+        for reg in self.r.iter_mut() {
+            *reg = regs.next().ok_or(())?
+        }
+        self.pc = regs.next().ok_or(())?;
+
+        if regs.next().is_some() {
+            return Err(());
+        }
+        Ok(())
```
daniel5151, line 68: nit: add a newline between the closing `}` and `Ok(())`

daniel5151:
Alright, it's merged in :tada: 

A couple of post-merge comments:

- If you end up adding support for the floating-point RISC-V extensions, take a quick look at this discussion https://github.com/daniel5151/gdbstub/pull/11#pullrequestreview-469853106 on the "Add x86_64 support" PR, and this https://github.com/daniel5151/gdbstub/issues/12 issue. Namely, the current `Arch` trait doesn't have any easy way to mix-and-match target features (e.g: [for RISC-V](https://sourceware.org/gdb/onlinedocs/gdb/RISC_002dV-Features.html#RISC_002dV-Features)) without a combinatoric increase in concrete impls.
    - For now, it'd probably be fine to just add a new `struct RiscvFpuRegs` + `RiscvWithFpu` arch implementation.
    - I'm no RISC-V expert, but a cursory glance at the arch shows that the other ISA extensions don't actually add any new registers, so having this small amount of redundancy shouldn't be too much of an issue.
- Unless there's some urgent need, I'd like to hold off on publishing a new version of `gdbstub` to `crates.io` until the floating point extension implementation is in (or I decide to release a new breaking version will all the misc. changes from the master branch). Is that alright?

thomashk0:
Thanks Daniel for the merge and the suggestions !

I planned to start with two different structs as you suggest. AFAIK, RISC-V extensions do not add general registers (they may introduce control register though (CSR), but I think GDB would query them through a single read register 'p' command). I think with and without FP cover 90% of RISC-V use cases !

I have no urgency to have it upstream on `crates.io`, I can stay on my fork as long as needed !

PR #22, opened 2020-08-30
Add 'read/write register' commands

thomashk0:
A proposal (probably to be reworked) API extension to support RSP "read register command".


daniel5151:
Nice, thanks for splitting up the PRs!

Like I mentioned in https://github.com/daniel5151/gdbstub/pull/21#issuecomment-683442358, the proposed API is leaking part of the underlying GDB protocol (namely, the `reg_number: usize` parameter), which is something `gdbstub` has tried to avoid.

Unfortunately, I can't think of a super simple way to easily hide this detail without a bit of plumbing and trait-rework 😬 

Here's what I'm thinking: extend the `Register` trait with an associated `RegId` type + `gdb_de/serialize_single` methods, and plumb that type-based register identification through into the user-facing `Target::read_register` API.

e.g: for RISC-V, the `RiscvCoreRegs` struct could have an associated `RiscvCoreRegsId` enum that looks something like:

```rust
enum RiscvCoreRegsId {
    R(usize), // shame there's no way to encode a u5 at the type level, ah well
    PC,
}
```

That way, the signature of `read_registers` could be changed to:

```rust
fn read_register(
     &mut self,
     dst: &mut [u8],
     reg: arch::riscv::reg::RiscvCoreRegsId,
 )
```

The end result is an API which enables end-users to uniquely identify a register _without_ having to search through the GDB source code to find the register numbers for their target.

I've got some other ideas on how it be possible to replace the `dst` parameter with some sort of direct reference into the `regs` struct (e.g: &mut usize), but I think it'll need some serious type-level shenanigans that can't really experiment with right now...

* * *

If all this sounds daunting, I don't blame you. 
`gdbstub` makes heavy use of Rust's powerful type system to eliminate as many runtime checks / conversions as possible. While this is definitely great from a performance / robustness point of view, it does tend to complicate the implementation quite a bit 😅 

If you want to take a crack at implementing some of these ideas (or proposing your own alternative!), I'd love to work with you to come up with something awesome 😄.

If that's not really something you'd be interested in at this time, then I'll go ahead and open an issue (something like "Add read/write single registers API"), and try to implement this feature at some point before releasing the next breaking release (which I've been hacking on intermittently for the past couple weeks).

thomashk0:
Pretty awesome explanation, thanks ! 

I'm definitely interested to work on something along those lines. Actually its a nice opportunity to play with some advanced Rust :) Hopefully I will have some time to work on it in the next days.

daniel5151:
Hey, good to hear! If you ever want to throw some ideas around, feel free to shoot me an email / DM me on Discord (daniel5151#4520).

Additionally, I maaaaaaay have nerd-sniped myself while working on a way to remove the `dst` parameter :sweat_smile:

Turns out that while it's possible to pull-off, it ends up being quite ugly. Here's a (very detailed) Rust Playground link with my experimentation:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=3f7b726e9dbf8ee2455f06c30ce093de

daniel5151:
By the way, I wanted to draw your attention to the `trait-based-opt-features` branch I've been working on.

It's a massive API overhaul which uses a pretty nifty type-system + compiler optimization trick to emulate optional trait methods at compile time (along with a bunch of other useful properties). It should make the `Target` API significantly more robust for end users, with the added benefit of simplifying a lot of the `GdbStub` internals that handle `OptResult` and the `MaybeUnimpl`.

Thankfully, the majority of the changes in `trait-based-opt-features` are localized to the `gdbstub_impl` and `target` modules, with the rest of the repo (notably `arch`) being left almost completely untouched.

I bring this up since there's a good chance you'll get a nasty looking merge conflict once the `trait-based-opt-features` is merged to master. That said, since the meat of this feature will likely be contained within the `arch` module, it'll only be the surface level plumbing that'll have to be adjusted. 

So, uh, sorry about that!

thomashk0:
Thanks for notifying !

I read you code snippet, pretty interesting (it blew my mind the first 5 min :p). If I understood correctly, you try to tie the register identifier with their storage, so that we can have different storage types for different identifiers?

I pushed my (probably naive) solution to the PR. Here are the important points:
- There is a new a `RegId` type in `Register` trait: adding this imposed to add a default type in all `arch` implementations (default type in trait is an unstable feature). There is also a new type `RegIdNotImplemented`, that can be used as a default for other `arch`.
- the trait Target gained a new 'read_register' function, with prototype `read_register(&mut self, Arch::Registers::RegId reg_id, &mut [u8]) -> ...`

A few remarks:
- The register type for a given register ID is not exposed at all (we just pass `&mut [u8]`)
- I added a `GdbSerializable` trait (that can serve both `RegId` and `Registers`), not sure if that is a good idea, leave me know

Does the approach seems good to you? I will take a look at PR #24 and try to adapt.

src/arch/traits.rs
```diff
@@ -13,6 +27,9 @@ use crate::internal::BeBytes;
 // TODO: add (optional?) trait methods for reading/writing specific register
 // (via it's GDB index)
 pub trait Registers: Default {
+    /// Register identifier for addressing single registers
+    type RegId: GdbSerializable + Default;
```
daniel5151, line 31: Can't say I'm a fan of having a `Default` bound here. I understand why you've added it (from an implementation point of view), but there isn't really such thing as a "default register identifier" at the protocol level...

src/arch/traits.rs
```diff
@@ -22,6 +39,15 @@ pub trait Registers: Default {
     fn gdb_deserialize(&mut self, bytes: &[u8]) -> Result<(), ()>;
 }
 
+/// Data that can be serialized according to GDB protocol.
+pub trait GdbSerializable {
```
daniel5151, line 43: To keep things DRY, you could just make `Registers` a supertrait of `GdbSerializable` (i.e: `trait Registers: Default + GdbSerializable`). After all, why have two sets of methods that do the exact same thing :smile:

daniel5151 (changes requested):
> I read you code snippet, pretty interesting (it blew my mind the first 5 min :p). If I understood correctly, you try to tie the register identifier with their storage, so that we can have different storage types for different identifiers?

Heh, glad you liked it! Pretty sneaky, eg? And yeah, your analysis was spot on.

* * *

I threw a couple of general comments in the review. Some may or may not be relevant depending on your thoughts on this next bit...

> The register type for a given register ID is not exposed at all (we just pass `&mut [u8]`)

Yeah, that's fine. As I showed above, making it bulletproof at the type level seems really tricky :wink: 

> I added a GdbSerializable trait...

I think a big issue is that the current `gdb_deserialize` method isn't particularly well suited to be implemented on a `RegId` enum. It forces `RegId` to have a `Default` type, as there needs to be some destination to write data into...

I did think of another idea though: instead of having `RegId` implement `GdbSerializable`, why not do something a simple as:

```rust
trait RegId {
    /// Map raw GDB register numbers to a (RegId, register size). 
    /// Returns None if the register isn't available.
    fn from_raw_id(id: usize) -> Option<(Self, usize)>;
}

// for unsupported / unimplemented systems
impl RegId for () {
    fn from_raw_id(id: usize) -> Option<(Self, usize)> { None }
}

// and in the `GdbStub` implementation:
Command::p(cmd) => {
    let mut dst = [0u8; 16];
    let reg = <<T::Arch as Arch>::Registers as Registers>::RegId::from_raw_id(cmd.reg_id);
    let (reg_id, size) = match reg {
        Some(v) => v,
        None => return Ok(None), // GDB docs say to return nothing for unsupported query
    }
    let dst = &mut dst[0..size];
    // enforce size within the GDB stub, instead of relying on the target
    target.read_register(dst, reg_id).maybe_missing_impl()?;
    res.write_hex_buf(dst)?;
}
```

I think this approach has a lot of merit, as it avoids introducing any redundant traits or weird type-bounds. Thoughts?

src/arch/traits.rs
thomashk0: I dropped the `GdbSerializable` trait in the new iteration.

thomashk0:
Thank you so much, your feedbacks are very helpful !! 
I'm much more satisfied with this version ! While the commit touches many files, the changes are really minor.

src/arch/traits.rs
```diff
@@ -2,6 +2,20 @@ use num_traits::{Num, PrimInt, Unsigned};
 
 use crate::internal::BeBytes;
 
+/// Stub register identifier.
+///
+/// Use this type as `Register::RegId` if your target does not handle register identifier.
+#[derive(Default)]
+pub struct NotImplementedRegId;
```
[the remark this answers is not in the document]
thomashk0: Yeah it sounds better !

src/arch/riscv/reg/riscv.rs
```diff
@@ -1,6 +1,41 @@
-use crate::arch::Registers;
+use crate::arch::{Registers, GdbSerializable};
 use crate::internal::LeBytes;
 use num_traits::PrimInt;
+use crate::protocol::prelude::decode_hex;
+
+/// RISC-V Register identifier.
+pub enum RiscvRegId {
+    Unknown,
```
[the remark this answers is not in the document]
thomashk0: Removed, the `Unknown` was required to have a sane `Default` implementation.

src/arch/traits.rs
thomashk0: +1, I'm happy you found a workaround !

src/arch/traits.rs
```diff
@@ -13,6 +29,12 @@ use crate::internal::BeBytes;
 // TODO: add (optional?) trait methods for reading/writing specific register
 // (via it's GDB index)
 pub trait Registers: Default {
+    /// Register identifier for addressing single registers.
+    ///
+    /// If your target does not implement that feature, you can use `RegId = ()`
```
daniel5151, line 34: This doc comment needs to be reworded.

- Avoid using "you"/"your" when possible.
- The end-user target has nothing to do with `RegId`, it's up to the `Arch` author to include a valid RegID.
- Add a note mentioning that `()` is a _temporary_ measure, and that all newly added architectures are required to provide a register identifier
    - Additionally, it would be nice to add a brief note to nudge people towards opening a PR if they see `RegID = ()` :smile:

daniel5151 (changes requested):
Nice, that's looking pretty slick! 
I'm glad we could come up with something that's nice and clean!

I've left a couple of comments inline / below, but this looks just about ready to ship :ship: 

> I still plan to add a few things in the RISC-V RegId within a few days (mostly adding missing registers). Maybe we can wait this is done before merging the PR?

Yep, just let me know when you're ready to merge :+1: 

* * *

- Could you also implement the `write_register` functionality as well?
    - It shouldn't be too difficult, and would keep the API nice and symmetrical 
    - _Hint:_ When parsing the `P` packet, you'll want to follow a similar pattern to the `G` packet, where the hex string is decoded "in place" using the `decode_hex_buf` helper, with the resulting packet holding a `&'a [u8]` reference to the decoded buffer.
- If it's not too much trouble, could you update the `armv4t` example with an example implementation of `read/write_register`? 
    - The corresponding `RegID` structure should be self-evident (the `ArmCoreRegs` struct is very straightforward)
    - Running the example should be fairly straightforward (see [examples/armv4t/README.md](https://github.com/daniel5151/gdbstub/blob/master/examples/armv4t/README.md)), and i've even included a pre-built test binary (`test.elf`) with debug symbols. It uses a relative path for the source map, so as long as you run `gdb-multiarch` from the `test_bin` directory, it should Just Work
    - The code is very simple, so even if you're not familiar with the details of the ARM architecture, it shouldn't be too difficult to get working :)

thomashk0:
Hey! Good news its almost ready to ship :smile: 

I applied your suggestions and also added the write register command. I'm still not fully satisfied with "what should I return as error?". The doc do not say (as with 'p') that nothing means "unsupported". However, looking at the source code (https://github.com/bsdjhb/gdb/blob/master/gdb/remote.c#L8361), GDB seems to handle the unsupported case the same way as 'p' does... . BTW the error code if any is returned to 'P' is just ignored, anything that starts with E should be accepted...

Yeah, I can add the RegId quite easily for Armvt4 I think (speaking ARM all days @work :p). It is a great idea as it would allow to have tests for that new part of the API. What about doing that in a dedicated issue/PR?

src/gdbstub_impl/mod.rs
```diff
@@ -413,6 +413,15 @@ impl<T: Target, C: Connection> GdbStubImpl<T, C> {
                 target.read_register(reg_id, dst).maybe_missing_impl()?;
                 res.write_hex_buf(dst)?;
             }
+            Command::P(p) => {
+                let reg = <<T::Arch as Arch>::Registers as Registers>::RegId::from_raw_id(p.reg_id);
+                let (reg_id, _) = match reg {
+                    Some(v) => v,
+                    None => return Ok(None),
```
daniel5151, line 420: (see main review comment)

daniel5151 (changes requested):
Nice! The parsing / handler / API surface of the `write_registers` command LGTM!

I've got a few more comments and nits, but they're mostly just minor doc-related fixes. 

> I'm still not fully satisfied with "what should I return as error?"

Lets just play it safe and go off what the protocol docs say (even if the GDB client implementation doesn't seem to care). i.e: add a `res.write_str("E01")?` inside the `None` arm of the `P` packet handler.

> Yeah, I can add the RegId quite easily for Armvt4... What about doing that in a dedicated issue/PR?

Could we just lump it in with this PR? After all, we're already lumping in some RISC-V definitions, so why not add some ARM ones as well :stuck_out_tongue: Plus, it would be good to have a working example of how to use the new API alongside the implementation itself.

daniel5151:
By the way, I've decided to hold-off on merging #24 until after this PR gets in, so don't worry about dealing with any merge conflicts, I'll just resolve them on my end :smile:

thomashk0:
Oh, cool!
Hopefully, I will manage to free some time tomorrow to complete that :)

thomashk0:
Regarding the `armv4t` example, I only added the `write_register`. To see it in action : 
```console
> target remote 9001
> set $r2=0xDEADBEEF
> info register
```
I struggle to get gdb send single read register for this architecture, so I removed `read_register`...

examples/armv4t/gdb.rs
```diff
@@ -5,6 +5,20 @@ use gdbstub::{
 };
 
 use crate::emu::{Emu, Event};
+use core::convert::TryInto;
```
daniel5151, line 8: tiiiny style nit: `core::` imports should be specified at the top of the file, and separately from external crate imports. Similarly, external crate imports should be separate from internal crate imports (notice blank line separating the two?).

So, move the core import up, and the `gdbstub` import up alongside the other `gdbstub` imports

examples/armv4t/gdb.rs
```diff
@@ -73,6 +87,20 @@ impl Target for Emu {
         Ok(())
     }
 
+    fn write_register(
```
daniel5151, line 90: I know you said that `read_register` would never be executed, but could you nonetheless re-add it so that it shows what the code could look like for other architectures? After all, that's sort of the point of this example, as I'm willing to bet that most consumers of the library won't actually compile / run it, just reference the code :smile:

src/protocol/commands.rs
```diff
@@ -102,7 +102,8 @@ commands! {
     "T" => _t_upcase::T,
     "z" => _z::z,
     "Z" => _z_upcase::Z,
-
+    "p" => _p::p,
+    "P" => _p_upcase::P<'a>,
```
daniel5151, line 106: nit: these should be inserted into the list in alphabetical order.

daniel5151 (changes requested):
Hooray! :tada: 

I've done one last full-sweep through the PR, found the last few tiiiiny style nits, but after those are fixed, I'm all ready to smash that merge button!

Thanks again for all the work you've put into the PR. It's a great feature, and I hope you had fun messing about with Rust's associated types and generics :smile:

P.S: before merging my _massive_ API overhaul PR, I'll publish one more version of gdbstub `0.2` with this feature, so that you won't have to upgrade to the new API if you don't want to :wink:

thomashk0:
Hey Daniel, I hope this last update will be fine :)

Many thanks for you very helpful reviews, I learned so much in the process! I think I will continue to contribute small things in `gdbstub` at least the 64 bit version of RISC-V + proper FP support!

Having that in 0.2 would be cool and allow my project to directly depend on the `crate.io` version.

daniel5151 (approved):
Wonderful! Time to :ship: this bad boy!

Looking forward to any other contributions you might want to make!

And small heads up, I've made a _minor_ API change in the new version on crates.io, where the `resume` method takes a `Actions` iterator instead of a raw `&mut dyn impl Iterator<..>`. This is entirely surface level, and doesn't change any functionality. Thankfully, semver [technically allows](https://semver.org/#spec-item-4) versions <1.x.y to make breaking API changes at any time :yum:

Oh, and shameless plug, but I would recommend updating to `0.3` once I publish it, since aside from API improvements, it comes with a couple of internal protocol fixes + optimizations. If you're curious what the API looks like, check out the `examples` directory on the development branch :smile: 

Cheers!

daniel5151:
I've pushed version `0.2.2` up to crates.io which includes this PR :smile:

PR #23, opened 2020-09-02
Add i386 support

jamcleod:


src/arch/x86/reg/mod.rs
```diff
@@ -1,9 +1,90 @@
...
+    pub fioff: u32,
+    /// FPU operand segment
+    pub foseg: u32,
+    /// FPU operand offset
+    pub fooff: u32,
+    /// Floating-point opcode
+    pub fop: u32,
+}
+
+impl TryFrom<&[u8]> for X87FpuInternalRegs {
```
daniel5151, line 35: Instead of implementing `TryFrom<&[u8]>` and `X87FpuInternalRegs::write`, `X87FpuInternalRegs` should just implement `Registers`. After all, the `try_from` method is essentially just `gdb_deserialize()`, and the `write` method is just `gdb_serialize()` :sweat_smile:

daniel5151 (changes requested):
Aside from my one comment, this LGTM.

Oh, and run the code through `rustfmt` :wink:

src/arch/x86/reg/core32.rs
```diff
@@ -0,0 +1,132 @@
...
+
+        // xmm0 to xmm15
+        for xmm_reg in &self.xmm {
+            write_bytes!(&xmm_reg.to_le_bytes());
+        }
+
+        // mxcsr
+        write_bytes!(&self.mxcsr.to_le_bytes());
+
+        (0..4).for_each(|_| write_byte(None))
```
daniel5151, line 83: nit: I assume this is more padding?
Could you add a brief comment like the one in `X86_64CoreRegs`?

daniel5151 (changes requested):
Nice!

I found _oooone_ last nit to fix, but after that, it's ready to ship :ship:

src/arch/x86/reg/core32.rs
jamcleod: I definitely had a comment there at some point, must've done some rearranging at got it removed when I switching between my i386 and ppc branches...

daniel5151: > ppc branches

👀

daniel5151 (approved):

PR #24, opened 2020-09-03
(API Overhaul) Trait-based Protocol Extensions

daniel5151:
This is a pretty massive API overhaul which uses a pretty nifty type-system + compiler optimization trick to emulate optional trait methods at compile time (along with a bunch of other useful properties, such as compile-time trait method grouping / mutual exclusion, and improved performance). 

As far as I can tell, I'm the first one to stumble across this hack (and/or make such extensive use of it), so I've given it the working title of "Inlineable Dyn Extension Traits" (IDETs). I'm already working on a writeup that talks about this technique! See https://github.com/daniel5151/optional-trait-methods

Technical details aside, the end-result of this PR is to make the Target API significantly more robust for end users, along with simplifying a lot of the GdbStub internals that handle OptResult and the MaybeUnimpl.

Scope-wise, this PR doesn't touch any code outside of the `gdbstub_impl` and `target` modules. Notably, the `arch` module is left completely unchanged.


daniel5151:
(these changes have been merged into master, so I'm closing this draft PR)

Issue #20, opened 2020-08-27
Add support for `qOffsets` query


mchesser:
For targets that relocate an image before execution, the debug symbols in GDB will have the incorrect offsets. The `qOffsets` query allows GDB to query the stub for the text segment offset after relocation.

I implemented the functionality I needed here: https://github.com/mchesser/gdbstub/commit/db347e0d9e1056549aaa7517938c608f27a32305 however there appears to be two different methods of reporting the offset (I think one is relative, and one is absolute), and I'm not entirely sure if I missed anything in the implementation

daniel5151:
Hmm, googling `qOffsets` brought up an issue on the `rust-gdb-remote-protocol` project which had some interesting discussion: 
 https://github.com/luser/rust-gdb-remote-protocol/issues/11

It looks like the cleaner (and more modern?) way to address the underlying issue would be to implement [`qXfer:libraries:read`](https://sourceware.org/gdb/current/onlinedocs/gdb/General-Query-Packets.html#qXfer-library-list-read) (and/or, [`qXfer:memory-map:read`](https://sourceware.org/gdb/current/onlinedocs/gdb/General-Query-Packets.html#qXfer-memory-map-read)). Moreover, unless I misread something in the linked discussion, it looks like `gdb-server` doesn't even implement `qOffsets`!

If it's not too much trouble, would you be willing to experiment with implementing one/both of these packets, and seeing if they work as intended?

Working with XML might be a bit clunky, but it shouldn't be too tricky to get things working. Oh, and you may need to add some logic to the `qSupported` response (i.e: sending `;qXfer:libraries:read+` and/or `qXfer:memory-map:read`).

* * *

P.S: What sort of project are you using `gdbstub` with? Just curious 😄

mchesser:
`qXfer:libraries:read` should work fine for my use, I'll have a go at implemented it.

I know the gdb-stub in qemu does respond to `qOffsets` https://github.com/qemu/qemu/blob/25f6dc28a3a8dd231c2c092a0e65bd796353c769/gdbstub.c#L2076-L2089 --  However I looking through the git history it looks like that code was originally from about 14 years ago! so I'm happy to move the the more modern way of doing things. 

(My project is an emulator that exposes a simple (also emulated) Linux userland -- I previously had my own gdb stub implementation (the protocol is a bit of a mess isn't it?), but it was much more limited than your `gdbstub` so I'm trying to switch over).

daniel5151:
[...]

PR #30, opened 2020-09-24
Add support for `qOffsets` packet

mchesser:
Implement the `qOffsets` packet allowing the stub to report section offsets back to gdb so that debug symbols are resolved to the correct addresses if the target performs any relocations.

Partially addresses: #20 

----

I saw the comment, about in progress API changes -- if necessary, I'm happy to update my branch once the API changes are complete.


src/target/mod.rs
```diff
@@ -317,6 +317,11 @@ pub trait Target {
     fn monitor_cmd(&mut self) -> Option<ext::MonitorCmdOps<Self>> {
         None
     }
+
+    /// Handle `qOffsets` command
```
daniel5151, line 321: (see top level comment regarding docs)

src/target/ext/offsets.rs
```diff
@@ -0,0 +1,18 @@
+//! Handle the `qOffsets` command
+
+use crate::target::Target;
+
+/// Target Extension - Handle the `qOffsets` command
+pub trait OffsetsCmd: Target {
+    /// Handle the `qOffsets` command
```
daniel5151, line 7: (see top level comment regarding docs)

src/target/ext/offsets.rs
```diff
@@ -0,0 +1,18 @@
+//! Handle the `qOffsets` command
```
daniel5151, line 1: (see top level comment regarding docs)

src/gdbstub_impl/mod.rs
```diff
@@ -23,6 +23,32 @@ pub use error::GdbStubError;
 
 use GdbStubError as Error;
 
+/// Describes the offset the target loaded the image sections at so the target
+/// can notify GDB that it needs to adjust the addresses of symbols
+pub enum SectionOffsets<U> {
```
daniel5151, line 28: Lets move this struct into `crate::target::ext::offsets`, since it's only relevant for those implementing the `SectionOffsets` extension.

src/target/ext/mod.rs
```diff
@@ -8,6 +8,7 @@ use crate::target::Target;
 
 pub mod breakpoint;
 pub mod monitor;
+pub mod offsets;
```
daniel5151, line 11: Could we change the module name to `section_offsets`, and the `OffsetsCmd` trait to `SectionOffsets`? That would keep things consistent with the rest of the extensions. 

Ignore `monitor::MonitorCmd`, I've fixed that inconsistency on a local branch lol

src/target/ext/offsets.rs
```diff
@@ -0,0 +1,18 @@
...
+
+use crate::target::Target;
+
+/// Target Extension - Handle the `qOffsets` command
+pub trait OffsetsCmd: Target {
+    /// Handle the `qOffsets` command
+    ///
+    /// For some targets, sections may be relocated from their base address. As
```
daniel5151, line 9: Could you push these docs up to the module level? It would make it easier to understand what this extension does while browsing through the generated docs.

src/target/ext/mod.rs
daniel5151: Alternatively, maybe the names should be changed to something along the lines of `legacy_section_offsets`? 
This would tie-back to the fact that `qOffsets` seems to have been superseded by `qXfer:libraries:read`, and that it would be great to support the new approach as well.

Or possibly even `basic_section_offsets`, reflecting the fact that `qOffsets` only allows specifying text/data relocation.

Up to you really.

src/target/ext/offsets.rs
```diff
@@ -0,0 +1,18 @@
...
+    /// a result, the stub may need to tell GDB the final section addresses
+    /// to ensure that debug symbols are resolved correctly after relocation.
+    ///
+    /// Implementing this command allows the stub to report text, data, and bss
+    /// offsets to GDB.
```
daniel5151, line 14: Could you add a `/// _Note:_` which mentions that this feature uses the legacy `qOffsets` packet, and that it would be good to add support for the newer and more flexible `qXfer:libraries:read` packet instead (linking back to #20)? The new packet support relocating arbitrary sections (not just Text/Data/BSS).

daniel5151 (changes requested):
Awesome, this looks great!

The inline comments are mostly around naming / docs, but the core implementation is spot-on. Couldn't have written it better myself!

And don't worry about rebasing on-top of my local changes. I've been bike-shedding how the project's modules are laid out, so it wouldn't be a particularly easy rebase. I'll just integrate these changes myself when I get the chance.

* * *

Regarding the doc comments: While it's tempting to mention the underlying `qOffsets` packet directly in the docs, I think it'd be best to re-word the docs to talk about the provided functionality, rather than the underlying implementation.

e.g: instead of "Handle the `qOffsets` command", maybe it could be "Report the section offsets the target used when relocating the running binary" (or something similar).

After all, one of the guiding principles behind `gdbstub` is to abstract over the gnarly underlying GDB protocol, and provide a "friendly" API on-top of the ~40 year old protocol :smile:

src/target/ext/mod.rs
mchesser: I've kept the name as `section_offsets` because while there are still modern alternatives, I still think it is probably the easiest method of to handle simple relocation adjustment. It seems unlikely that GDB will deprecate the method any time soon since is the only method supported by QEMU.

mchesser:
So I was slightly wrong about the semantics of the different offset responses -- I've been working with "stripped" binaries that are missing section metadata and only have segment information available, so I misunderstood the different between section offsets and segment address. 

I also finally found the proper modern way of indicating the base address of the dynamic linker for System V binaries (where GDB is capable of extracting the `link_map` from memory). It appears there is another optional command that you can respond with `qXfer:auxv:read`. The auxiliary vector contains a entry (`AT_BASE`) that represents the exact address that I want to notify GDB of, you can see here: https://github.com/bminor/binutils-gdb/blob/9e790a80160676e7fd3fb8be6cf3c1c77d9ded81/gdb/solib-svr4.c#L2349 that it is used for obtaining the library list.

The commit where they added `qOffsets` support `lldb` is enlightening, looks like currently `qOffsets` is really only supported because that is what QEMU uses, everyone else responds with `auxv`.

I've tried adding some docs pointing readers to the correct information, but let me know what you think.

daniel5151 (changes requested):
Ahhhh, the plot thickens!
Thanks for drilling down into the details, I really appreciate it.

Could you update #20 with a quick summary of what you've discovered regarding `qXfer::auxv::read` and `qXfer::library::read`? This could be in the form of a new comment, or preferably, by updating the top-level comment (for better visibility).

In terms of code/docs, I've got a few more minor nits (gotta remember to add those periods!), but aside from those, the PR LGTM :+1:

src/target/ext/section_offsets.rs
```diff
@@ -0,0 +1,60 @@
...
+        text_seg: U,
+        /// The absolute address of the second segment which conventionally
+        /// contains modifiable data
+        data_seg: Option<U>,
+    },
+}
+
+/// Target Extension - section/segment relocation offsets from the target.
+///
+/// For some targets, sections may be relocated from their base address. As
```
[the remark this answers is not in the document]
mchesser: Yeah, I 100% sure about what you meant because of how the docs done for the `monitor` command, but I've updated this now.

src/gdbstub_impl/mod.rs
```diff
@@ -653,7 +627,7 @@ impl<T: Target, C: Connection> GdbStubImpl<T, C> {
 
                 if let Some(op) = target.section_offsets() {
                     match op.get_section_offsets().map_err(Error::TargetError)? {
-                        SectionOffsets::Relative { text, data, bss } => {
+                        ext::section_offsets::Offsets::Sections { text, data, bss } => {
```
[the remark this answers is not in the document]
mchesser: Yeah, this is a habit I've pickup because rust-analyzer doesn't support local imports yet.

src/target/mod.rs
```diff
@@ -318,7 +318,7 @@ pub trait Target {
         None
     }
 
-    /// Handle `qOffsets` command
+    /// Handle requests to get the target's current section (or segment) offsets
```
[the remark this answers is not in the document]
mchesser: Ah... periods -- this is kind of a stylistic thing I do, because I personally think it looks a bit noisy (particularly on struct fields), definitely looks better when rendered with rustdoc though.

daniel5151 (approved):

daniel5151:
Awesome work!
Thanks again for your contribution!  :tada: 

I've got a few more unmerged changes sitting around on my local machine (such as `target extended-remote` support and improved error handling), so once those are cleaned up + merged into master, I'll probably go ahead and publish `0.4.0` (possibly with a `-beta1` tag) crates.io. 

Cheers!

PR #33, opened 2020-10-26
specify cfg-if 0.1.10 or later

keiichiw:
`cargo build` fails if cfg-if is 0.1.9 or older.

It seems that [this change](https://github.com/alexcrichton/cfg-if/pull/28) is needed to compile [this part](https://github.com/daniel5151/gdbstub/blob/da716bdd7b/src/protocol/console_output.rs#L43-L48).


daniel5151:
Good catch, thanks!

Issue #29, opened 2020-09-17
Implement missing arch-specific `RegId` implementations


daniel5151:
# Overview

#22 added support for register-level read/writes, and introduced a new `RegId` associated type to the existing `Registers` trait. This associated type is used to translate raw GDB register ids (i.e: a arch-dependent `usize`) into a structured human-readable enum identifying the register.

e.g:

```rust
/// 32-bit ARM core register identifier.
#[derive(Debug, Clone, Copy)]
pub enum ArmCoreRegId {
    /// General purpose registers (R0-R12)
    Gpr(u8),
    /// Stack Pointer (R13)
    Sp,
    /// Link Register (R14)
    Lr,
    /// Program Counter (R15)
    Pc,
    /// Floating point registers (F0-F7)
    Fpr(u8),
    /// Floating point status
    Fps,
    /// Current Program Status Register (cpsr)
    Cpsr,
}

impl RegId for ArmCoreRegId {
    fn from_raw_id(id: usize) -> Option<(Self, usize)> {
        match id {
            0..=12 => Some((Self::Gpr(id as u8), 4)),
            13 => Some((Self::Sp, 4)),
            14 => Some((Self::Lr, 4)),
            15 => Some((Self::Pc, 4)),
            16..=23 => Some((Self::Fpr(id as u8), 4)),
            25 => Some((Self::Cpsr, 4)),
            _ => None,
        }
    }
}
```

Unfortunately, this feature was only implemented _after_ several contributors had already upstreamed their `Arch` implementations. 
As such, the `RawRegId` type was implemented as a "stop gap" measure to allow users to manually implement `read/write_registers`, even if it meant looking up the raw GDB register numbers themselves.

Please refer to the docs for more details on `RawRegId`, but at a glance:

```rust
#[derive(Debug)]
pub struct RawRegId(pub usize);

impl RegId for RawRegId {
    // simply pass-through the raw register ID +
    fn from_raw_id(id: usize) -> Option<(Self, usize)> {
        Some((RawRegId(id), 256 / 8))
    }
}
```

# Action Items

While all future `arch` implementations will be required to implement a proper `RegId`, the older `arch` implementations will have to be manually updated to support the new `RegId` system.

At the moment, these are the `arch` implementations that are in the main repo.
Those with a [x] have a valid `RegId`, while those with a [ ] are still using `RawRegId`:

- arm
  - [x] `Armv4t`
- mips
  - [ ] `Mips` / `Mips64`
- msp430
[...]

PR #34, opened 2020-10-26
implement `RegId` for x86/x86_64

keiichiw:
This is one of action items in issue #29.

### DIFF
--- a/src/arch/x86/mod.rs
+++ b/src/arch/x86/mod.rs
@@ -10,7 +10,7 @@ pub mod reg;
 /// Check out the [module level docs](../index.html#whats-with-regidimpl) for
 /// more info about the `RegIdImpl` type parameter.
 #[allow(non_camel_case_types)]
-pub enum X86_64_SSE<RegIdImpl: RegId> {
+pub enum X86_64_SSE<RegIdImpl: RegId = reg::id::X86_64CoreRegId> {
     #[doc(hidden)]
     _Marker(core::marker::PhantomData<RegIdImpl>),
 }
--- a/src/arch/x86/mod.rs
+++ b/src/arch/x86/mod.rs
@@ -32,7 +32,7 @@ impl<RegIdImpl: RegId> Arch for X86_64_SSE<RegIdImpl> {
 /// Check out the [module level docs](../index.html#whats-with-regidimpl) for
 /// more info about the `RegIdImpl` type parameter.
 #[allow(non_camel_case_types)]
-pub enum X86_SSE<RegIdImpl: RegId> {
+pub enum X86_SSE<RegIdImpl: RegId = reg::id::X86CoreRegId> {
     #[doc(hidden)]
     _Marker(core::marker::PhantomData<RegIdImpl>),
 }
--- a/src/arch/x86/reg/id.rs
+++ b/src/arch/x86/reg/id.rs
@@ -1,5 +1,148 @@
-// TODO: Add proper `RegId` implementation. See [issue #29](https://github.com/daniel5151/gdbstub/issues/29)
-// pub enum X86RegId {}
+use crate::arch::RegId;
+
+/// FPU register identifier.
+#[derive(Debug, Clone, Copy)]
+pub enum X87FpuInternalRegId {
+    /// Floating-point control register
+    Fctrl,
+    /// Floating-point status register
+    Fstat,
+    /// Tag word
+    Ftag,
+    /// FPU instruction pointer segment
+    Fiseg,
+    /// FPU intstruction pointer offset
+    Fioff,
+    /// FPU operand segment
+    Foseg,
+    /// FPU operand offset
+    Fooff,
+    /// Floating-point opcode
+    Fop,
+}
+
+impl X87FpuInternalRegId {
+    fn from_u8(val: u8) -> Option<Self> {
+        use self::X87FpuInternalRegId::*;
+
+        let r = match val {
+            0 => Fctrl,
+            1 => Fstat,
+            2 => Ftag,
+            3 => Fiseg,
+            4 => Fioff,
+            5 => Foseg,
+            6 => Fooff,
+            7 => Fop,
+            _ => return None,
+        };
+        Some(r)
+    }
+}
+
+/// 32-bit x86 core + SSE register identifier.
+///
+/// Source: https://github.com/bminor/binutils-gdb/blob/master/gdb/features/i386/32bit-core.xml
+/// Additionally: https://github.com/bminor/binutils-gdb/blob/master/gdb/features/i386/32bit-sse.xml
+#[derive(Debug, Clone, Copy)]
+#[non_exhaustive]
+pub enum X86CoreRegId {
+    /// Accumulator
+    Eax,
+    /// Count register
+    Ecx,
+    /// Data register
+    Edx,
+    /// Base register
+    Ebx,
+    /// Stack pointer
+    Esp,
+    /// Base pointer
+    Ebp,
+    /// Source index
+    Esi,
+    /// Destination index
+    Edi,
+    /// Instruction pointer
+    Eip,
+    /// Status register
+    Eflags,
+    /// Segment registers: CS, SS, DS, ES, FS, GS
+    Segment(u8),
+    /// FPU registers: ST0 through ST7
+    St(u8),
+    /// FPU internal registers
+    Fpu(X87FpuInternalRegId),
+    /// SIMD Registers: XMM0 through XMM7
+    Xmm(u8),
+    /// SSE Status/Control Register
+    Mxcsr,
+}
+
+impl RegId for X86CoreRegId {
+    fn from_raw_id(id: usize) -> Option<(Self, usize)> {
+        use self::X86CoreRegId::*;
+
+        let r = match id {
+            0 => (Eax, 4),
+            1 => (Ecx, 4),
+            2 => (Edx, 4),
+            3 => (Ebx, 4),
+            4 => (Esp, 4),
+            5 => (Ebp, 4),
+            6 => (Esi, 4),
+            7 => (Edi, 4),
+            8 => (Eip, 4),
+            9 => (Eflags, 4),
+            10..=15 => (Segment(id as u8 - 10), 4),
+            16..=23 => (St(id as u8 - 16), 10),
+            24..=31 => match X87FpuInternalRegId::from_u8(id as u8 - 24) {
+                Some(r) => (Fpu(r), 4),
+                None => unreachable!(),
+            },
+            32..=39 => (Xmm(id as u8 - 32), 16),
+            40 => (Mxcsr, 4),
+            _ => return None,
+        };
+        Some(r)
+    }
+}
 
 // TODO: Add proper `RegId` implementation. See [issue #29](https://github.com/daniel5151/gdbstub/issues/29)
 // pub enum X86_64RegId {}
+
+#[cfg(test)]
+mod tests {
+    use crate::arch::traits::RegId;
+    use crate::arch::traits::Registers;
+
+    /// Compare the following two values which are expected to be the same:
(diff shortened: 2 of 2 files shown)


src/arch/x86/reg/id.rs
```diff
@@ -1,5 +1,48 @@
...
-// pub enum X86_64RegId {}
+/// 64-bit x86 core + SSE register identifier.
+///
+/// Source: https://github.com/bminor/binutils-gdb/blob/master/gdb/features/i386/64bit-core.xml
+/// Additionally: https://github.com/bminor/binutils-gdb/blob/master/gdb/features/i386/64bit-sse.xml
+#[derive(Debug, Clone, Copy)]
+#[non_exhaustive]
+pub enum X86_64CoreRegId {
+    /// General purpose registers: RAX, RBX, RCX, RDX, RSI, RDI, RBP, RSP, r8-r15
+    Gpr(u8),
```
keiichiw, line 14: I also considered to have `Rax`, `Rbx`, etc as different enum variants but chose this design to align with the design of `X86_64CoreRegs` for now.
Which design do you prefer?

keiichiw:
@daniel5151 Please take a look.
FYI, I'm working to use this crate in [crosvm](https://chromium-review.googlesource.com/c/chromiumos/platform/crosvm/+/2440221). (Yeah, I know you worked on crosvm before!)

src/arch/x86/reg/id.rs
daniel5151: Yep, the current design seems reasonable. 

I've typically deferred these sorts of details to the PR author, since they're more familiar with the target architecture. That said, I'm always willing to change things around (assuming you've got a good reason to prefer one over the other)

src/arch/x86/reg/id.rs
```diff
@@ -1,5 +1,50 @@
...
+    Rip,
+    /// Status register
+    Eflags,
+    /// Segment registers: CS, SS, DS, ES, FS, GS
+    Segment(u8),
+    /// FPU registers: ST0 through ST7
+    St(u8),
+    /// FPU internal registers: FCTRL, FSTAT, FTAG, FISEG, FIOFF, FOSEG, FOOFF,
+    /// FOP
+    Fpu(u8),
```
daniel5151, line 26: Instead of a `u8`, could we instead have a separate `X87FpuInternalRegId` enum which corresponds to the `X87FpuInternalRegs`?

I don't think it'd need it's own `impl RegId`, as the implementation might differ between `x86` and `x86_64`.

src/arch/x86/reg/id.rs
```diff
@@ -1,5 +1,50 @@
...
+        use crate::arch::x86::reg::id::X86_64CoreRegId::*;
+
+        let r = match id {
+            0..=15 => (Gpr(id as u8), 8),
+            16 => (Rip, 4),
+            17 => (Eflags, 8),
+            18..=23 => (Segment(id as u8), 4),
+            24..=31 => (St(id as u8), 10),
+            32..=40 => (Fpu(id as u8), 4),
+            41..=55 => (Xmm(id as u8), 16),
```
daniel5151, line 44: These should be `(id - <offset>) as u8`

src/arch/x86/reg/id.rs
```diff
@@ -1,5 +1,50 @@
+use crate::arch::RegId;
+
 // TODO: Add proper `RegId` implementation. See [issue #29](https://github.com/daniel5151/gdbstub/issues/29)
 // pub enum X86RegId {}
```
daniel5151, line 4: If it's not too much trouble, could you also implement the X86RegId as well?

Feel free to say no -- it's just that the two implementations aught to be fairly similar (I'd imagine), and it'd be nice to get both of them out of the way at the same time.

src/arch/x86/reg/id.rs
```diff
@@ -1,5 +1,50 @@
...
+    Gpr(u8),
+    /// Instruction pointer
+    Rip,
+    /// Status register
+    Eflags,
+    /// Segment registers: CS, SS, DS, ES, FS, GS
+    Segment(u8),
+    /// FPU registers: ST0 through ST7
+    St(u8),
+    /// FPU internal registers: FCTRL, FSTAT, FTAG, FISEG, FIOFF, FOSEG, FOOFF,
```
daniel5151, line 24: nit: could you put the newline between `registers:` and `FCTRL`?

That way, there won't be a single straggling `FOP` on the next line :)

src/arch/x86/mod.rs
```diff
@@ -10,15 +10,12 @@ pub mod reg;
...
 /// more info about the `RegIdImpl` type parameter.
 #[allow(non_camel_case_types)]
-pub enum X86_64_SSE<RegIdImpl: RegId> {
-    #[doc(hidden)]
-    _Marker(core::marker::PhantomData<RegIdImpl>),
-}
+pub enum X86_64_SSE {}
 
-impl<RegIdImpl: RegId> Arch for X86_64_SSE<RegIdImpl> {
+impl Arch for X86_64_SSE {
```
daniel5151, line 15: Please leave the generic parameter in-place. Instead, set `RegIdImpl = reg::id::X86_64CoreRegId` by default.

As per the last paragraph in https://docs.rs/gdbstub/0.4.0/gdbstub/arch/index.html#whats-with-regidimpl, this will allow me to make a single breaking change after all missing `RegId` implementations have been resolved.

src/arch/x86/reg/id.rs
```diff
@@ -1,5 +1,50 @@
...
-// TODO: Add proper `RegId` implementation. See [issue #29](https://github.com/daniel5151/gdbstub/issues/29)
-// pub enum X86_64RegId {}
+/// 64-bit x86 core + SSE register identifier.
+///
+/// Source: https://github.com/bminor/binutils-gdb/blob/master/gdb/features/i386/64bit-core.xml
+/// Additionally: https://github.com/bminor/binutils-gdb/blob/master/gdb/features/i386/64bit-sse.xml
+#[derive(Debug, Clone, Copy)]
+#[non_exhaustive]
+pub enum X86_64CoreRegId {
+    /// General purpose registers: RAX, RBX, RCX, RDX, RSI, RDI, RBP, RSP,
```
daniel5151, line 13: nit: could you put the newline between `registers:` and `RAX`?

That way, there won't be a single straggling `r8-r15` on the next line :)

daniel5151 (changes requested):
Thank you for the PR!

I have a couple of comments, but all in all, this LGTM :+1: 

* * *

I won't lie, I'm pretty excited to see that my library might be integrated into `crosvm`!
I had a blast working on crosvm as an intern, and I'm glad to be able to contribute back to it again (albeit indirectly :wink:)

src/arch/x86/reg/id.rs
keiichiw: The order of general purpose registers seems not to be well defined. 
It looks inconsistent even in the Intel SDM. For example, RSI comes before RDI in some paragraphs, but it sometimes comes after RDI in other parts:( 
That's why I had considered to have different variants for these registers. Actually, `X86CoreRegs` does so. ([crosvm](https://chromium.googlesource.com/chromiumos/platform/crosvm/+/d4ca29b5d9920de4ec387b8c89a6d30a6d4fca79/hypervisor/src/x86_64.rs#451) also does so.)

But, it's probably a different topic. The current design is also reasonable, as it simply follows the order defined in the GDB's XML file and the order is documented well. So, let me leave this design as is.

src/arch/x86/reg/id.rs
keiichiw: Done. Instead of implementing `RegId`, I used `FromPrimitive` to convert an integer to an enum value.

src/arch/x86/reg/id.rs
keiichiw: Oops! Fixed.

src/arch/x86/reg/id.rs
keiichiw: Sure!

src/arch/x86/reg/id.rs
keiichiw: Removed the list of register names because I defined a separate enum for FPU registers.

src/arch/x86/mod.rs
keiichiw: Sorry, I missed that part in the document. Fixed.

src/arch/x86/reg/id.rs
keiichiw: Done

keiichiw:
Please take a look again. I addressed your comments. Also, I added tests.

src/arch/x86/reg/id.rs
daniel5151: Ahh, classic x86...

Personally, I'm now leaning towards breaking out RAX, RBX, ..., RSP into their own variants (and now that I think about it, the Segment registers as well). That said, I think it's fine if we leave the design as-is, as it matches the existing `x86[_64]CoreRegs` structure.

I'll probably open a tracking issue to break-up the GPRs + SRs into their own struct fields / variants, though that should be something I can take care of myself at some point closer to the release of `0.5`

src/arch/x86/reg/id.rs
```diff
@@ -1,5 +1,182 @@
...
+    Fpu(X87FpuInternalRegId),
+    /// SIMD Registers: XMM0 through XMM7
+    Xmm(u8),
+    /// SSE Status/Control Register
+    Mxcsr,
+}
+
+impl RegId for X86CoreRegId {
+    fn from_raw_id(id: usize) -> Option<(Self, usize)> {
+        use crate::arch::x86::reg::id::X86CoreRegId::*;
```
daniel5151, line 66:
```suggestion
        use self::X86CoreRegId::*;
```

src/arch/x86/reg/id.rs
```diff
@@ -1,5 +1,182 @@
...
+            40..=55 => (Xmm(id as u8 - 40), 16),
+            56 => (Mxcsr, 4),
+            _ => return None,
+        };
+        Some(r)
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
daniel5151, line 144: More tests are always appreciated! Thanks!

src/arch/x86/reg/id.rs
keiichiw: Sounds good!
The register names are really confusing. Actually, when I was implementing the GDB feature for crosvm, I mistakenly swapped the order of RBP and RSP, and wasted a whole day to debug it😓 
So, I'm really supportive of breaking GPRs + SRs into their own variants.

Cargo.toml
```diff
@@ -16,6 +16,7 @@ categories = ["development-tools::debugging", "embedded", "emulators", "network-
 cfg-if = "0.1"
 log = "0.4"
 managed = { version = "0.8", default-features = false }
+num-derive = { version = "0.3", default-features = false }
```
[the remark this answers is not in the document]
keiichiw: It makes sense. Done.

src/arch/x86/reg/id.rs
keiichiw: Sure! I tried to make this test architecture-independent. So, I hope we'll be able to have similar tests for other targets.

Unfortunately, the current test can't be applied to `ArmCoreRegId` as is, though. It's because floating point registers are ignored in `gdb_serialize`.

keiichiw:
Thanks for the comments and suggestions! I addressed them 😄

daniel5151 (approved):
I'll go ahead and push out a minor version sometime today that includes your two PRs.

Fantastic work, thanks again!

src/arch/x86/reg/id.rs
daniel5151: I've gone ahead and opened the tracking issue at #35 

If you're bored, feel free to open a PR for it yourself, otherwise, I'll try to get around to it at some arbitrary point in the future :smile:

PR #37, opened 2020-11-05
Exclude test object files from package

keiichiw:
Exclude test object files such as `/examples/armv4t/test.o` when running `cargo package`. I guess these data are not expected to be distributed to all gdbstub users.
Also, excluding those files makes it easier to use this crate on the portage package manager, as it complains about prebuilt binaries and requires [some hack ](https://chromium-review.googlesource.com/c/chromiumos/overlays/chromiumos-overlay/+/2507271/4/dev-rust/gdbstub/gdbstub-0.4.0.ebuild#29).


daniel5151:
Ahh, portage. Now that's a name that bring back some ~good~ memories...

Sure, this change seems perfectly reasonable. I've also gone ahead and removed the `.o` files from the repo, as those shouldn't have been committed in the first place (not too sure how they slipped past my radar, but hey, better late than never ¯\\\_(ツ)_/¯)

daniel5151 (approved):

keiichiw:
Thanks for the quick approval! Are you planning to release a new version with this PR? Then, I can remove a hack in our ChromeOS code base:)
Please feel free to say no if you don't want to do.

daniel5151:
Sure :+1: 

Version `0.4.2` should be published on crates.io now.

keiichiw:
Thanks a lot! I could just remove some workaround from [the ebuild](https://chromium-review.googlesource.com/c/chromiumos/overlays/chromiumos-overlay/+/2507271).

PR #38, opened 2021-01-08
Implement `RegId` for Mips and MSP430

starfleetcadet75:
This PR addresses the missing arch implementations (MIPS32/64, MSP430, and PPC) for `RegId` mentioned in issue #29. It also adds the MIPS DSP register struct.


src/arch/ppc/reg/id.rs
```diff
@@ -1,2 +1,87 @@
-// TODO: Add proper `RegId` implementation. See [issue #29](https://github.com/daniel5151/gdbstub/issues/29)
-// pub enum PowerPc32RegId {}
+use crate::arch::RegId;
+
+/// 32-bit PowerPC register identifier.
```
daniel5151, line 3: Please include a link to the XML file these defns are from.

daniel5151 (changes requested):
Hey, thanks for the PR!

There are a couple issues that need to be fixed before we can merge, but the core definitions seem fine.

I do have one big question - while I'm fairly confident that you've tested the Mips RegId (given that you contributed the initial `Mips` arch to `gdbstub`), have you actually tested the msp430 or Power PC definitions?

While it'd certainly be nice to close #29, the reason I opened the issue in the first place + implemented this somewhat janky-looking `RegIdImpl` approach was because I didn't feel comfortable adding definitions that haven't been tested-working in at least one project. 

I've taken the same approach with every other architecture-specific piece of functionality in `gdbstub`. Case in point: the only arch I've actually added myself was `armv4t` - every other arch has been a community contribution!

While I appreciate the effort, if the msp430 and Power PC `RegId`s haven't actually been tested, then I'd want to omit them from this PR, and focus on just the Mips `RegId`s. I don't mind leaving #29 open until someone with more msp430 / Power PC experience comes along to fill in the gaps, since the issue doesn't really affect the broader codebase / end user experience all that much.

src/arch/mips/mod.rs
```diff
@@ -9,16 +9,25 @@ pub mod reg;
...
-pub enum Mips<RegIdImpl: RegId> {
+pub enum Mips<RegIdImpl: RegId = reg::id::MipsRegId<u32>> {
     #[doc(hidden)]
     _Marker(core::marker::PhantomData<RegIdImpl>),
 }
 
 /// Implements `Arch` for 64-bit MIPS.
-///
-/// Check out the [module level docs](../index.html#whats-with-regidimpl) for
-/// more info about the `RegIdImpl` type parameter.
```
daniel5151, line 20: this bit of the comment is still relevant

src/arch/mips/reg/mips.rs
```diff
@@ -1,5 +1,6 @@
 use crate::arch::Registers;
 use crate::internal::LeBytes;
+use core::convert::TryInto;
 
 use num_traits::PrimInt;
```
daniel5151, line 5: nit: I just realized the imports should be shuffled around a bit:

```suggestion
use core::convert::TryInto;

use num_traits::PrimInt;

use crate::arch::Registers;
use crate::internal::LeBytes;
```

src/arch/mips/reg/mips.rs
```diff
@@ -99,11 +136,12 @@ where
     fn gdb_deserialize(&mut self, bytes: &[u8]) -> Result<(), ()> {
         let ptrsize = core::mem::size_of::<U>();
 
-        // ensure bytes.chunks_exact(ptrsize) won't panic
-        if bytes.len() % ptrsize != 0 {
+        // Ensure bytes contains enough data for all 72 registers
+        if bytes.len() < ptrsize * 72 {
```
daniel5151, line 140: These two checks aren't equivalent.

For example, if `bytes.len() == ptrsize * 72 + 1`, then `chunks_exact` will panic.

You can leave this condition in, but do not remove the other one.

src/arch/mips/reg/mips.rs
```diff
@@ -136,10 +174,85 @@ where
...
+    }
+
+    fn gdb_deserialize(&mut self, bytes: &[u8]) -> Result<(), ()> {
+        // Deserialize the core registers first
+        self.core.gdb_deserialize(bytes)?;
+
+        // Ensure bytes contains enough data for all 79 registers of target-width
+        // and the dspctl register which is always 4 bytes
+        let ptrsize = core::mem::size_of::<U>();
+        if bytes.len() < (ptrsize * 79) + 4 {
```
daniel5151, line 223: Same comment - this check is not sufficient to ensure that `chunks_exact` won't panic.

You can leave it in, but you still need the other check as well.

src/arch/mips/reg/mod.rs
```diff
@@ -6,5 +6,4 @@ pub mod id;
 mod mips;
 
 pub use mips::MipsCoreRegs;
-pub use mips::MipsCp0Regs;
-pub use mips::MipsFpuRegs;
```
daniel5151, line 10: This is a semver breaking change - these must remain pub.

src/arch/mips/mod.rs
```diff
@@ -9,16 +9,25 @@ pub mod reg;
...
-pub enum Mips64<RegIdImpl: RegId> {
+pub enum Mips64<RegIdImpl: RegId = reg::id::MipsRegId<u64>> {
+    #[doc(hidden)]
+    _Marker(core::marker::PhantomData<RegIdImpl>),
+}
+
+/// Implements `Arch` for 32-bit MIPS with the DSP feature enabled.
+pub enum MipsWithDsp<RegIdImpl: RegId = reg::id::MipsRegId<u32>> {
```
daniel5151, line 24: Since this is a new type, the whole `RegIdImpl` rigmarole is unnecessary.

src/arch/mips/mod.rs
```diff
@@ -9,16 +9,25 @@ pub mod reg;
...
+    #[doc(hidden)]
+    _Marker(core::marker::PhantomData<RegIdImpl>),
+}
+
+/// Implements `Arch` for 64-bit MIPS with the DSP feature enabled.
+pub enum Mips64WithDsp<RegIdImpl: RegId = reg::id::MipsRegId<u64>> {
```
daniel5151, line 30: ditto

daniel5151 (changes requested):
Nice, good stuff!

Just a few more nits, but aside from that this looks pretty much good to merge.

Given that you've removed the PowerPC work, could you update the PR title?

src/arch/mips/reg/mips.rs
```diff
@@ -22,6 +22,8 @@ pub struct MipsCoreRegs<U> {
     pub cp0: MipsCp0Regs<U>,
     /// FPU registers
     pub fpu: MipsFpuRegs<U>,
+    /// DSP registers
+    pub dsp: MipsDspRegs<U>,
```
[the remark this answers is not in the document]
daniel5151: (leaving this comment unresolved so that I can easily find it in the future, if I ever need to reference it again)

starfleetcadet75:
I've separated the MIPS DSP struct from the core registers, which made me realize interestingly enough that `dspctl` is the only MIPS register that doesn't change size when going from 32 to 64 bit. I have tested the Mips defs with my emulator, however I don't have anything to test PPC with so I went ahead and pulled the PPC stuff out for now. I do have some experience with msp430, unfortunately it looks like GDB doesn't have a nice XML description for it so I linked what seems to be the closest thing in the gdb source.

src/arch/mips/reg/mips.rs
starfleetcadet75: The issue I'm seeing with `if bytes.len() % ptrsize != 0 {` is when U is a u64 there will be 79 registers of size 64 and one register of size 32 (dspctl), thus this check will fail. It looks like [chunks_exact](https://doc.rust-lang.org/core/slice/struct.ChunksExact.html) actually doesn't panic when there are extra elements, it should still split the bytes up until the end with a possible remainder of 4 bytes.

daniel5151: Ahh, yep, you're right.

Well, in any case, I decided to see if the optimizer was smart enough to get rid of any panics, and yep, sure enough it seems to have done the trick:

https://godbolt.org/z/8zsKf9

So yeah, this check can be left as-is.

daniel5151 (approved):
LGTM :+1: 

Thanks for the contribution!

I'll go ahead and publish `gdbstub 0.4.3` with these changes shortly.

PR #39, opened 2021-02-16
Fix trait terminology

pheki:
OptExt is a subtrait of Target, not the opposite. See https://doc.rust-lang.org/reference/items/traits.html#supertraits

Also thanks for the read 😄


daniel5151:
Ah, nice catch! Thanks!

Issue #47, opened 2021-04-13
Support reverse-execution `b` commands


DrChat:
Add support for the GDB reverse set of commands.
This corresponds to `ReverseContinue+` and `ReverseStep+` in the `qSupported` features.

This will likely warrant a new set of ops, such as a `ReverseExecutionOps`.

daniel5151:
A brief skim through the GDB docs implies that this might be as simple as adding support for the `bc` and `bs` packets, and sending back the appropriate `qSupported` features. Unfortunately, while simple to implement at the protocol level, it might be surprisingly tricky to "nicely" implement this functionality in `gdbstub`:

1. How does `bs` know which thread to reverse step?
    -  `bc` and `bs` seem to mirror the _deprecated_ `c` and `s` packets, which do not package "execution context" (i.e: thread id) as part of the packets themselves. `c` and `s` relied on a separate `H` packet to communicate that execution context information.
    -  From my cursory reading of the GDB docs, it doesn't seem to explicitly document how `bc` and `bs` know which thread to operate on (admittedly, I don't think it really matters for `bc`, since it'll just run the _whole system_ in reverse). I'll have to see what the reference implementation is doing...
2. How to expose reverse execution functionality at the API level?
    - _Note:_ This should actually be fairly straightforward in `gdbstub 0.5` (assuming the current WIP resume API stays as-is), and would likely mirror how the optional range-steping API looks like. The difficulty would be in exposing this functionality in a _non-breaking manner_ as part of `0.4.x`.

In any case, this is certainly a feature worth supporting, and is something I'll keep an eye on when I find more time to hack away at `gdbstub`. In the meantime, feel free to dig a bit deeper into the question/challenges I outlined above, and if you'd like, consider whipping up a PR that implements the feature (preferably targeting the `dev/0.5` branch).

DrChat:
Perfect - thanks for the technical dive-in! I'll be spending some time looking at this as part of my current project, since reverse-execution support is desirable.

DrChat:
One question is whether or not `ResumeAction` can be extended.
It looks like for your range-stepping implementation, you did not extend the enum, so multi-thread targets would have to handle it themselves.

It'd be more ideal to extend the enum instead, though I'm not familiar enough with the language to understand whether or not it's feasible or convenient to do so.

daniel5151:
> It'd be more ideal to extend the enum instead

[...]

PR #48, opened 2021-04-21
Reverse execution

DrChat:
This is a pull request for reverse continue/step support for gdbstub.
This corresponds to the `bc` / `bs` packets, or the `rs` / `rc` commands in GDB.

### DIFF
--- a/examples/armv4t/gdb/mod.rs
+++ b/examples/armv4t/gdb/mod.rs
@@ -2,7 +2,10 @@ use core::convert::TryInto;
 
 use armv4t_emu::{reg, Memory};
 use gdbstub::target;
-use gdbstub::target::ext::base::singlethread::{ResumeAction, SingleThreadOps, StopReason};
+use gdbstub::target::ext::base::singlethread::{
+    ResumeAction, SingleThreadOps, SingleThreadReverseContOps, SingleThreadReverseStepOps,
+    StopReason,
+};
 use gdbstub::target::ext::breakpoints::WatchKind;
 use gdbstub::target::{Target, TargetError, TargetResult};
 use gdbstub_arch::arm::reg::id::ArmCoreRegId;
@@ -157,6 +160,14 @@ impl SingleThreadOps for Emu {
         Some(self)
     }
 
+    fn support_reverse_cont(&mut self) -> Option<SingleThreadReverseContOps<Self>> {
+        Some(self)
+    }
+
+    fn support_reverse_step(&mut self) -> Option<SingleThreadReverseStepOps<Self>> {
+        Some(self)
+    }
+
     fn support_resume_range_step(
         &mut self,
     ) -> Option<target::ext::base::singlethread::SingleThreadRangeSteppingOps<Self>> {
@@ -196,6 +207,32 @@ impl target::ext::base::SingleRegisterAccess<()> for Emu {
         } else {
             Err(().into())
         }
+    }
+}
+
+impl target::ext::base::singlethread::SingleThreadReverseCont for Emu {
+    fn reverse_cont(
+        &mut self,
+        check_gdb_interrupt: &mut dyn FnMut() -> bool,
+    ) -> Result<StopReason<u32>, Self::Error> {
+        // FIXME: actually implement reverse step
+        eprintln!(
+            "FIXME: Not actually reverse-continuing. Performing forwards continue instead..."
+        );
+        self.resume(ResumeAction::Continue, check_gdb_interrupt)
+    }
+}
+
+impl target::ext::base::singlethread::SingleThreadReverseStep for Emu {
+    fn reverse_step(
+        &mut self,
+        check_gdb_interrupt: &mut dyn FnMut() -> bool,
+    ) -> Result<StopReason<u32>, Self::Error> {
+        // FIXME: actually implement reverse step
+        eprintln!(
+            "FIXME: Not actually reverse-stepping. Performing single forwards step instead..."
+        );
+        self.resume(ResumeAction::Step, check_gdb_interrupt)
     }
 }
 
--- a/src/gdbstub_impl/ext/base.rs
+++ b/src/gdbstub_impl/ext/base.rs
@@ -4,7 +4,7 @@ use crate::protocol::commands::ext::Base;
 use crate::arch::{Arch, Registers};
 use crate::protocol::{IdKind, SpecificIdKind, SpecificThreadId};
 use crate::target::ext::base::multithread::ThreadStopReason;
-use crate::target::ext::base::{BaseOps, ResumeAction};
+use crate::target::ext::base::{BaseOps, ReplayLogPosition, ResumeAction};
 use crate::{FAKE_PID, SINGLE_THREAD_TID};
 
 impl<T: Target, C: Connection> GdbStubImpl<T, C> {
@@ -27,6 +27,25 @@ impl<T: Target, C: Connection> GdbStubImpl<T, C> {
                 res.write_str(";vContSupported+")?;
                 res.write_str(";multiprocess+")?;
                 res.write_str(";QStartNoAckMode+")?;
+
+                let (reverse_cont, reverse_step) = match target.base_ops() {
+                    BaseOps::MultiThread(ops) => (
+                        ops.support_reverse_cont().is_some(),
+                        ops.support_reverse_step().is_some(),
+                    ),
+                    BaseOps::SingleThread(ops) => (
+                        ops.support_reverse_cont().is_some(),
+                        ops.support_reverse_step().is_some(),
+                    ),
+                };
+
+                if reverse_cont {
+                    res.write_str(";ReverseCont+")?;
+                }
+
+                if reverse_step {
+                    res.write_str(";ReverseStep+")?;
+                }
 
                 if let Some(ops) = target.extended_mode() {
                     if ops.configure_aslr().is_some() {
@@ -557,7 +576,7 @@ impl<T: Target, C: Connection> GdbStubImpl<T, C> {
                 BaseOps::MultiThread(ops) => Self::do_vcont_multi_thread(ops, res, &actions)?,
             };
 
-            match self.finish_vcont(res, target, stop_reason)? {
+            match self.finish_exec(res, target, stop_reason)? {
                 Some(status) => break Ok(status),
                 None => continue,
             }
@@ -565,7 +584,7 @@ impl<T: Target, C: Connection> GdbStubImpl<T, C> {
     }
 
     #[inline(always)]
-    fn finish_vcont(
+    pub(super) fn finish_exec(
         &mut self,
         res: &mut ResponseWriter<C>,
         _target: &mut T,
@@ -619,6 +638,17 @@ impl<T: Target, C: Connection> GdbStubImpl<T, C> {
                 res.write_str(";")?;
                 HandlerStatus::Handled
             }
+            ThreadStopReason::ReplayLog(pos) => {
+                res.write_str("T05")?;
+
+                res.write_str("replaylog:")?;
+                res.write_str(match pos {
+                    ReplayLogPosition::Begin => "begin",
+                    ReplayLogPosition::End => "end",
+                })?;
+
+                HandlerStatus::Handled
+            }
         };
 
         Ok(Some(status))
@@ -640,6 +670,7 @@ impl<U> From<StopReason<U>> for ThreadStopReason<U> {
                 addr,
             },
             StopReason::Signal(sig) => ThreadStopReason::Signal(sig),
+            StopReason::ReplayLog(pos) => ThreadStopReason::ReplayLog(pos),
         }
     }
 }
--- a/src/gdbstub_impl/ext/reverse_exec.rs
+++ b/src/gdbstub_impl/ext/reverse_exec.rs
@@ -1,0 +1,141 @@
+use super::prelude::*;
+use crate::protocol::commands::ext::{ReverseCont, ReverseStep};
+
+use crate::arch::Arch;
+use crate::protocol::SpecificIdKind;
+use crate::target::ext::base::multithread::{MultiThreadReverseCont, MultiThreadReverseStep};
+use crate::target::ext::base::singlethread::{SingleThreadReverseCont, SingleThreadReverseStep};
+use crate::target::ext::base::BaseOps;
+
+enum ReverseContOps<'a, A: Arch, E> {
+    SingleThread(&'a mut dyn SingleThreadReverseCont<Arch = A, Error = E>),
+    MultiThread(&'a mut dyn MultiThreadReverseCont<Arch = A, Error = E>),
+}
+
+enum ReverseStepOps<'a, A: Arch, E> {
+    SingleThread(&'a mut dyn SingleThreadReverseStep<Arch = A, Error = E>),
+    MultiThread(&'a mut dyn MultiThreadReverseStep<Arch = A, Error = E>),
+}
+
+impl<T: Target, C: Connection> GdbStubImpl<T, C> {
+    pub(crate) fn handle_reverse_cont(
+        &mut self,
+        res: &mut ResponseWriter<C>,
+        target: &mut T,
+        command: ReverseCont,
+    ) -> Result<HandlerStatus, Error<T::Error, C::Error>> {
+        // Resolve the reverse-continue operations. Error out if the target does not
+        // support it.
+        let ops = match target.base_ops() {
+            BaseOps::MultiThread(ops) => match ops.support_reverse_cont() {
+                Some(ops) => ReverseContOps::MultiThread(ops),
+                None => return Ok(HandlerStatus::Handled),
+            },
+            BaseOps::SingleThread(ops) => match ops.support_reverse_cont() {
+                Some(ops) => ReverseContOps::SingleThread(ops),
+                None => return Ok(HandlerStatus::Handled),
+            },
+        };
+
+        let handler_status = match command {
+            ReverseCont::bc(_) => {
+                // FIXME: look into why this isn't being optimized out.
+                crate::__dead_code_marker!("bc", "impl");
+
+                // FIXME: This block is duplicated from the vCont code.
+                let mut err = Ok(());
+                let mut check_gdb_interrupt = || match res.as_conn().peek() {
+                    Ok(Some(0x03)) => true, // 0x03 is the interrupt byte
+                    Ok(Some(_)) => false,   // it's nothing that can't wait...
+                    Ok(None) => false,
+                    Err(e) => {
+                        err = Err(Error::ConnectionRead(e));
+                        true // break ASAP if a connection error occurred
+                    }
+                };
+
+                let stop_reason = match ops {
+                    ReverseContOps::MultiThread(ops) => ops
+                        .reverse_cont(&mut check_gdb_interrupt)
+                        .map_err(Error::TargetError)?,
+                    ReverseContOps::SingleThread(ops) => ops
+                        .reverse_cont(&mut check_gdb_interrupt)
+                        .map_err(Error::TargetError)?
+                        .into(),
+                };
+
+                err?;
+
+                // FIXME: properly handle None case
+                self.finish_exec(res, target, stop_reason)?
+                    .ok_or(Error::PacketUnexpected)?
+            }
+        };
+
+        Ok(handler_status)
+    }
+
+    // FIXME: De-duplicate with above code?
+    pub(crate) fn handle_reverse_step(
+        &mut self,
+        res: &mut ResponseWriter<C>,
+        target: &mut T,
+        command: ReverseStep,
+    ) -> Result<HandlerStatus, Error<T::Error, C::Error>> {
+        // Resolve the reverse-step operations. Error out if the target does not
+        // support it.
+        let ops = match target.base_ops() {
+            BaseOps::MultiThread(ops) => match ops.support_reverse_step() {
+                Some(ops) => ReverseStepOps::MultiThread(ops),
+                None => return Ok(HandlerStatus::Handled),
+            },
+            BaseOps::SingleThread(ops) => match ops.support_reverse_step() {
+                Some(ops) => ReverseStepOps::SingleThread(ops),
+                None => return Ok(HandlerStatus::Handled),
+            },
+        };
+
+        let handler_status = match command {
+            ReverseStep::bs(_) => {
+                // FIXME: look into why this isn't being optimized out.
+                crate::__dead_code_marker!("bs", "impl");
+
+                let tid = match self.current_resume_tid {
+                    // NOTE: Can't single-step all cores.
+                    SpecificIdKind::All => return Err(Error::PacketUnexpected),
+                    SpecificIdKind::WithId(tid) => tid,
+                };
+
+                // FIXME: This block is duplicated from the vCont code.
+                let mut err = Ok(());
+                let mut check_gdb_interrupt = || match res.as_conn().peek() {
+                    Ok(Some(0x03)) => true, // 0x03 is the interrupt byte
+                    Ok(Some(_)) => false,   // it's nothing that can't wait...
+                    Ok(None) => false,
+                    Err(e) => {
+                        err = Err(Error::ConnectionRead(e));
+                        true // break ASAP if a connection error occurred
+                    }
+                };
+
+                let stop_reason = match ops {
+                    ReverseStepOps::MultiThread(ops) => ops
+                        .reverse_step(tid, &mut check_gdb_interrupt)
+                        .map_err(Error::TargetError)?,
+                    ReverseStepOps::SingleThread(ops) => ops
+                        .reverse_step(&mut check_gdb_interrupt)
+                        .map_err(Error::TargetError)?
+                        .into(),
+                };
+
+                err?;
+
+                // FIXME: properly handle None case
+                self.finish_exec(res, target, stop_reason)?
+                    .ok_or(Error::PacketUnexpected)?
+            }
+        };
+
+        Ok(handler_status)
+    }
+}
--- a/src/protocol/commands.rs
+++ b/src/protocol/commands.rs
@@ -79,6 +79,8 @@ macro_rules! commands {
                 trait Hack {
                     fn base(&mut self) -> Option<()>;
                     fn single_register_access(&mut self) -> Option<()>;
+                    fn reverse_step(&mut self) -> Option<()>;
+                    fn reverse_cont(&mut self) -> Option<()>;
                 }
 
                 impl<T: Target> Hack for T {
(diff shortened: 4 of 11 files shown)


daniel5151 (commented):
I've left a few inline nits and comments, but on the whole the implementation is looking good.

That said, now that I see this implementation shaping up... does it even make sense to model reverse step/continue as "virtual" vCont actions? After all, it will never be the case that a GDB client would request that some threads execute backwards while simultaneously requesting other threads to execute forwards - that would be nonsensical.

I'd initially thought reverse execution could be implemented similarly to how range-stepping was implemented, but now that I think about it, reverse execution is actually it's own "class" of resume actions...

As such, I think it might make sense to swap-out the surface-level API to something like the following:

```rust
pub enum ReverseResumeAction {
    Step(Tid),
    // Should Continue also specify a Tid? 
    // Probably not, right?
    Continue,
}

// and associated SingleThread variant
pub trait MultiThreadReverseExecution: Target + MultiThreadOps {
    fn resume(
        &mut self,
        reverse_action: ReverseResumeAction<Tid>,
        check_gdb_interrupt: &mut dyn FnMut() -> bool,
    ) -> Result<ThreadStopReason<<Self::Arch as Arch>::Usize>, Self::Error>;
}

define_ext!(MultiThreadReverseExecutionOps, MultiThreadReverseExecution);
```

After all, reverse execution doesn't actually have the capability to specify multiple resume actions per-thread. All it has are standalone `bc` and `bs` packets (until someone decides to implement some kind of `vReverseCont` packet).

Let me know what you think.

DrChat:
👍 I really appreciate the quick review! Let me address your comments.

DrChat:
I've pushed a set of changes to address your review comments.
Most importantly, I went ahead and refactored the code so it doesn't flow through `vCont` anymore, so reverse-stepping (and reverse-continue) are their own execution functions.

src/target/ext/base/multithread.rs
```diff
@@ -186,6 +193,23 @@ pub trait MultiThreadOps: Target {
...
 
+/// Target Extension - [Reverse execution] for multi threaded targets.
+///
+/// Reverse execution allows the target to run backwards up until there is
+/// no more execution history.
+///
+/// [Reverse execution]: https://sourceware.org/gdb/current/onlinedocs/gdb/Reverse-Execution.html
+pub trait MultiThreadReverseStep: Target + MultiThreadOps {
+    /// Reverse-step the specified [Tid] by one instruction.
+    fn reverse_step(
```
daniel5151, line 204: Is there a reason you're splitting `reverse_step` and `reverse_continue` into two separate functions (as opposed to using that `ReverseResumeAction` enum I mentioned earlier)? Or have you just not gotten around to updating it yet?

daniel5151 (commented):
Thanks for integrating my feedback so quickly, I think this approach is definitely better.

I left a few more nits and comments, but overall, I think this PR is on the right track.

src/target/ext/base/multithread.rs
DrChat: It's just because the GDB protocol also has them split into `ReverseContinue+;` and `ReverseStep+;` features. Do you think I should combine them anyways?

daniel5151: Ohhh, true...

If possible, could you look in to whether the two features are coupled, as https://github.com/daniel5151/gdbstub/issues/47#issuecomment-822215249 suggests? This might be an unfortunate case of the protocol implying one thing, but the GDB implementation doing another... 

In any case, I agree that the API should reflect the underlying protocol features, but it would be good to check if there's a undocumented dependency between the two, as that's something I would most certainly want to include in the documentation.

mchesser: I checked this again, and I'm now fairly sure you can separate out the features -- Not sure why I previously thought they were dependent on each other.

src/protocol/commands.rs
```diff
@@ -163,6 +163,8 @@ commands! {
         "T" => _t_upcase::T,
         "vCont" => _vCont::vCont<'a>,
         "vKill" => _vKill::vKill,
+        "bc" => _bc::bc,
```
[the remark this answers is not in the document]
DrChat: I don't see an example of doing this to extend `MultiThreadOps` or `SingleThreadOps`?

src/gdbstub_impl/ext/base.rs
```diff
@@ -300,6 +306,48 @@ impl<T: Target, C: Connection> GdbStubImpl<T, C> {
...
+            }
+
+            Base::bs(_) => {
+                let tid = match self.current_resume_tid {
+                    // N.B: Can't single-step all cores.
+                    SpecificIdKind::All => return Err(Error::PacketUnexpected),
+                    SpecificIdKind::WithId(tid) => tid,
+                };
+
+                // FIXME: This block is duplicated from the vCont code.
```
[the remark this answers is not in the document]
DrChat: 👍

src/gdbstub_impl/ext/base.rs
```diff
@@ -551,26 +599,21 @@ impl<T: Target, C: Connection> GdbStubImpl<T, C> {
         target: &mut T,
         actions: crate::protocol::commands::_vCont::Actions,
     ) -> Result<HandlerStatus, Error<T::Error, C::Error>> {
-        loop {
-            let stop_reason = match target.base_ops() {
-                BaseOps::SingleThread(ops) => Self::do_vcont_single_thread(ops, res, &actions)?,
-                BaseOps::MultiThread(ops) => Self::do_vcont_multi_thread(ops, res, &actions)?,
-            };
+        let stop_reason = match target.base_ops() {
```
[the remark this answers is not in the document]
DrChat: Ah okay, yeah I thought it was vestigal logic from something :)

DrChat:
Workflow approvals? That's a new one. I wonder if it's related to bitcoin mining.

daniel5151:
Oh wow, yeah, would you look at that. Fascinating...
I'd wager to bet that it's bitcoin related.

DrChat:
Haha, yeah. As always thanks for the quick review!
I've addressed your comments and pushed more changes.

src/protocol/commands.rs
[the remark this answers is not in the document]
daniel5151: Ah, I see...

To cut a long story short, my `commands!` macro assumes that packets correspond directly to top-level IDETs, which isn't really the case here. As I implement more of the protocol, I'm starting to realize that isn't always the case... but I haven't had the chance to rework the implementation just yet.

My current approach is a bit jank, as it uses a locally-scoped extension trait to define "pseudo" top-level IDETs specifically in the context of packet parsing. Take a look at the `single_register_access` method in the following code to see what I'm talking about: https://github.com/daniel5151/gdbstub/blob/54683d1/src/protocol/commands.rs#L76-L96

Assuming I didn't miss anything, you _should_ be able to define a `reverse_execution` method that mirrors the existing `single_register_access` method, which would then enable breaking these two packets out into a separate `reverse_execution` command group.

daniel5151:
Cool, thanks for addressing the changes. Now that I've outlined how to split up the `bc` and `bs` packet parsing code, you should be on track to wrap up the implementation.

I'll also take this chance to mention that prior to merging any new feature, I ask that contributors provide an implementation of the feature as part of the in-tree `armv4t` example.

Obviously, I don't actually expect you to implement working time-travel debugging in that shoddy little emulator (unless you have time to kill and don't mind hacking together some quick-and-dirty snapshotting lol), so a simple "degenerate" implementation that performs a regular 'ol single-step / continue should suffice. Just be sure to include something like `eprintln!("NOTE: reverse-{step/continue}, isn't implemented. Performing regular {step/continue} instead")` to validate that the function is indeed being called.

Oh, and of course, make sure to flesh out the docs more when you get the chance 😄

daniel5151:
_sigh_, it looks there's no way to perma-approve workflow runs on a PR.

I guess I'll just click the button whenever I get the chance then.

DrChat:
Yeah, that's annoying.
Don't worry about the last commit though, I just pushed it so I can switch computers.

DrChat:
Okay - I pushed another iteration up and addressed more of your comments.

P.S. is there a way to run those github lint hooks locally? Something like typing in `cargo check` and having it do everything?

daniel5151:
I'll take a look in a bit...

Those hooks aren't anything special. IIRC they're just:

```bash
cargo test --all --features=std

cargo clippy --all --examples --features=std -- -D warnings
cargo fmt --all -- --check

cargo fmt --manifest-path example_no_std/Cargo.toml -- --check
cargo clippy --manifest-path example_no_std/Cargo.toml
```

src/gdbstub_impl/ext/base.rs
```diff
@@ -615,7 +571,7 @@ impl<T: Target, C: Connection> GdbStubImpl<T, C> {
     }
 
     #[inline(always)]
-    fn finish_exec(
+    pub(crate) fn finish_exec(
```
daniel5151, line 574: Any reason why this needs to be `pub(crate)`?

src/gdbstub_impl/ext/reverse_exec.rs
```diff
@@ -0,0 +1,106 @@
+use super::prelude::*;
+use crate::protocol::{
+    commands::ext::{ReverseCont, ReverseStep},
+    SpecificIdKind,
+};
+
+use crate::target::ext::base::BaseOps;
```
daniel5151, line 7: nit: For consistency, can you match the import order +style used in other modules?

e.g: https://github.com/daniel5151/gdbstub/blob/3084b31/src/gdbstub_impl/ext/base.rs#L1-L8

- group 1 should have the prelude + the specific ext(s) this file implements handler(s) for
- group 2 has the rest of the crate-level imports

src/gdbstub_impl/ext/reverse_exec.rs
```diff
@@ -0,0 +1,106 @@
...
+                    }
+
+                    BaseOps::SingleThread(_ops) => {
+                        return Err(Error::PacketUnexpected);
+                    }
+                };
+
+                err?;
+
+                // FIXME: Unnecessary unwrap
```
daniel5151, line 47: yeah... unwraps are a strict no-no in the `gdbstub` codebase. Unless you've double checked the generated asm and confirmed that the panic branch has been optimized out, this needs to be handled without `panic!`ing

The "proper" fix in the future would be to introduce some similar looping logic, but for now, it should be fine to simply `.ok_or(Error::PacketUnexpected)`. It's not _strictly_ correct, but it'll do for now.

Please add a `// FIXME: properly handle None case` comment as well.

src/gdbstub_impl/ext/reverse_exec.rs
```diff
@@ -0,0 +1,106 @@
...
```
daniel5151, line 99: ditto

src/gdbstub_impl/ext/reverse_exec.rs
```diff
@@ -0,0 +1,106 @@
...
+                };
+
+                let stop_reason = match target.base_ops() {
+                    BaseOps::MultiThread(ops) => {
+                        if let Some(ops) = ops.support_reverse_step() {
+                            ops.reverse_step(tid, &mut check_gdb_interrupt)
+                                .map_err(Error::TargetError)?
+                        } else {
+                            return Err(Error::PacketUnexpected);
+                        }
```
daniel5151, line 89: Instead of doing this "late" operation check, you should lift this up to the top of the function, matching how other extensions handle this: https://github.com/daniel5151/gdbstub/blob/3084b311c20a41d966271be023ea39c13870a9cf/src/gdbstub_impl/ext/monitor_cmd.rs#L13-L16

One minor road-bump I forsee is that `let ops` will have to be an enum of either the single threaded or multi threaded ops. You'll probably have to manually define a locally scoped enum similar to the following:

```rust
enum ReverseStepOps { 
    SingleThread(<single threaded ops type>), 
    MultiThread(<multi threaded ops type>) 
}
```

src/gdbstub_impl/ext/reverse_exec.rs
```diff
@@ -0,0 +1,106 @@
...
+                    }
+                };
+
+                let stop_reason = match target.base_ops() {
+                    BaseOps::MultiThread(ops) => {
+                        if let Some(ops) = ops.support_reverse_step() {
+                            ops.reverse_step(tid, &mut check_gdb_interrupt)
+                                .map_err(Error::TargetError)?
+                        } else {
+                            return Err(Error::PacketUnexpected);
```
daniel5151, line 88: `return Err(Error::PacketUnexpected)` is _not_ correct here, and should be `return Ok(HandlerStatus::Handled)`. The GDB protocol allows clients to send packets the target doesn't understand / implement, and the correct response is to simply ignore the packet and send back nothing.

daniel5151 (changes requested):
Left some comments. Looking forward to seeing the single thread implementations as well :)

src/gdbstub_impl/ext/base.rs
DrChat: I'm reusing this code in `reverse_exec.rs` to send back a thread stop reason.

daniel5151: Oh, maybe I have my crate visibility rules mixed up, but I thought this would have worked without it...
If not, all good.

That said, can you check if `pub(super)` works? If it does, it'd be a bit more appropriate.

src/gdbstub_impl/ext/reverse_exec.rs
daniel5151: Oh, and I should mention that this isn't just a style nit, and has a tangible binary size impact.

The compiler is smart enough to inline `handle_X` methods (since they are only ever called once), and by having the early ops guard, the optimizer is able to completely discard the body of the function in cases where the target hasn't implemented the IDET.

DrChat: Really? Huh, very interesting. I'm toying around with the `__dead_code_marker` stuff now (which is pretty nifty), and it still looks like Rust isn't eliminating my new handlers for `bc` and `bs`.
I'll push the new iteration in a bit so you can take a look.

src/gdbstub_impl/ext/base.rs
DrChat: Indeed that works! I'll swap it over.

src/gdbstub_impl/ext/reverse_exec.rs
daniel5151: Yeah, I've been ricing `gdbstub` pretty hard, often leaning on godbolt and `cargo asm` to make sure things are in good shape.

I'd also recommend running `example_no_std/check_size.sh` to make sure you're not accidentally ballooning the base-configuration code size.

daniel5151:
(I'm not sure if you're waiting for more review comments from me. I'd prefer it if you just barreled forwards with the implementation, and we can iterate on the nearly-finished product)

DrChat:
Ah, sure. I guess I'm enjoying the timely reviews too much 😅
I'll finish up my set of changes and give it to you for review. Still learning a lot about Rust, but I've got a pretty good handle on the language now.

src/target/ext/base/singlethread.rs
```diff
@@ -98,6 +112,39 @@ pub trait SingleThreadOps: Target {
...
+/// Reverse stepping allows the target to run backwards by one step.
+/// Usually this corresponds to a single instruction, but it can be any metric
+/// defined by the target.
+///
+/// [Reverse stepping]: https://sourceware.org/gdb/current/onlinedocs/gdb/Reverse-Execution.html
+pub trait SingleThreadReverseStep: Target + SingleThreadOps {
+    /// Reverse-step the specified [`Tid`].
+    fn reverse_step(
+        &mut self,
+        check_gdb_interrupt: &mut dyn FnMut() -> bool,
```
DrChat, line 142: Unsure if this is strictly necessary, though it's possible that these operations may take some time.

DrChat:
Final set of changes pushed, this time it should be in a state ready to merge.

src/target/ext/base/singlethread.rs
daniel5151: It would be weird not to have it.

daniel5151 (changes requested):

src/gdbstub_impl/ext/base.rs
```diff
@@ -28,6 +28,28 @@ impl<T: Target, C: Connection> GdbStubImpl<T, C> {
...
+                        if ops.support_reverse_cont().is_some() {
+                            res.write_str(";ReverseCont+")?;
+                        }
+
+                        if ops.support_reverse_step().is_some() {
+                            res.write_str(";ReverseStep+")?;
+                        }
+                    }
+                }
+
```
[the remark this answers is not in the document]
DrChat: Perfect! Yeah, the logic I wrote was kind of nasty.

src/gdbstub_impl/ext/reverse_exec.rs
```diff
@@ -0,0 +1,141 @@
...
+
+            BaseOps::SingleThread(ops) => match ops.support_reverse_cont() {
+                Some(ops) => ReverseContOps::SingleThread(ops),
+                None => return Err(Error::PacketUnexpected),
+            }
+        };
+
+        let handler_status = match command {
+            ReverseCont::bc(_) => {
+                crate::__dead_code_marker!("bc", "impl");
```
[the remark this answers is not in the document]
DrChat: Analysis seems to suggest that this is not being optimized out on my Windows builds.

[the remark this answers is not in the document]
daniel5151: hmm, nor is it being optimized out on my linux builds...
I wouldn't stress too much over it. I can figure out why myself later.

examples/armv4t/gdb/mod.rs
```diff
@@ -199,6 +210,27 @@ impl target::ext::base::SingleRegisterAccess<()> for Emu {
     }
 }
 
+impl target::ext::base::singlethread::SingleThreadReverseCont for Emu {
+    fn reverse_cont(
+        &mut self,
+        check_gdb_interrupt: &mut dyn FnMut() -> bool,
+    ) -> Result<StopReason<u32>, Self::Error> {
+        // NOTE: GDB doesn't actually care if we go backwards :)
+        eprintln!("FIXME: Not actually reverse-continuing.");
```
daniel5151, line 219:
```suggestion
        // FIXME: actually implement reverse continue
        eprintln!("NOTE: Not actually reverse-continuing! Performing single forwards step instead...");
```

examples/armv4t/gdb/mod.rs
```diff
@@ -199,6 +210,27 @@ impl target::ext::base::SingleRegisterAccess<()> for Emu {
...
+        Ok(self.resume(ResumeAction::Step, check_gdb_interrupt)?)
+    }
+}
+
+impl target::ext::base::singlethread::SingleThreadReverseStep for Emu {
+    fn reverse_step(
+        &mut self,
+        check_gdb_interrupt: &mut dyn FnMut() -> bool,
+    ) -> Result<StopReason<u32>, Self::Error> {
+        eprintln!("FIXME: Not actually reverse-stepping.");
```
daniel5151, line 229:
```suggestion
        // FIXME: actually implement reverse step
        eprintln!("NOTE: Not actually reverse-stepping! Performing single forwards step instead...");
```

src/gdbstub_impl/ext/reverse_exec.rs
```diff
@@ -0,0 +1,140 @@
...
+                        err = Err(Error::ConnectionRead(e));
+                        true // break ASAP if a connection error occurred
+                    }
+                };
+
+                let stop_reason = match ops {
+                    ReverseStepOps::MultiThread(ops) => ops
+                        .reverse_step(tid, &mut check_gdb_interrupt)
+                        .map_err(Error::TargetError)?,
+
```
daniel5151, line 123: nit: remove blank line

src/target/ext/base/multithread.rs
```diff
@@ -186,6 +200,39 @@ pub trait MultiThreadOps: Target {
...
+        check_gdb_interrupt: &mut dyn FnMut() -> bool,
+    ) -> Result<ThreadStopReason<<Self::Arch as Arch>::Usize>, Self::Error>;
+}
+
+define_ext!(MultiThreadReverseContOps, MultiThreadReverseCont);
+
+/// Target Extension - [Reverse stepping] for multi threaded targets.
+///
+/// Reverse stepping allows the target to run backwards by one step.
+/// Usually this corresponds to a single instruction, but it can be any metric.
```
daniel5151, line 222: missed this line

src/target/ext/base/singlethread.rs
```diff
@@ -98,6 +112,38 @@ pub trait SingleThreadOps: Target {
...
+    ) -> Result<StopReason<<Self::Arch as Arch>::Usize>, Self::Error>;
+}
+
+define_ext!(SingleThreadReverseContOps, SingleThreadReverseCont);
+
+/// Target Extension - [Reverse stepping] for single threaded targets.
+///
+/// Reverse stepping allows the target to run backwards by one step.
+/// Usually this corresponds to a single instruction, but it can be any metric.
```
daniel5151, line 134: ...and this one too

daniel5151 (changes requested):

daniel5151:
btw: whenever you publish changes that you're happy with, please ask for a re-review so I get notified :)

src/gdbstub_impl/ext/reverse_exec.rs
DrChat: Oops! Is there a rustfmt lint for this by any chance?

src/gdbstub_impl/ext/reverse_exec.rs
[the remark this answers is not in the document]
daniel5151: can you add a `// FIXME: look into why this isn't being optimized out`?

src/gdbstub_impl/ext/reverse_exec.rs
```diff
@@ -0,0 +1,139 @@
...
+            },
+            BaseOps::SingleThread(ops) => match ops.support_reverse_step() {
+                Some(ops) => ReverseStepOps::SingleThread(ops),
+                None => return Ok(HandlerStatus::Handled),
+            },
+        };
+
+        let handler_status = match command {
+            ReverseStep::bs(_) => {
+                crate::__dead_code_marker!("bs", "impl");
```
daniel5151, line 99: ditto on this one

src/target/ext/base/multithread.rs
```diff
@@ -186,6 +200,39 @@ pub trait MultiThreadOps: Target {
...
+        check_gdb_interrupt: &mut dyn FnMut() -> bool,
+    ) -> Result<ThreadStopReason<<Self::Arch as Arch>::Usize>, Self::Error>;
+}
+
+define_ext!(MultiThreadReverseContOps, MultiThreadReverseCont);
+
+/// Target Extension - [Reverse stepping] for multi threaded targets.
+///
+/// Reverse stepping allows the target to run backwards by one step.
+/// Usually this corresponds to a single instruction.
```
daniel5151, line 222: ...just remove this line entirely.
The term `step` is entirely sufficient.

src/target/ext/base/singlethread.rs
```diff
@@ -98,6 +112,38 @@ pub trait SingleThreadOps: Target {
...
+        check_gdb_interrupt: &mut dyn FnMut() -> bool,
+    ) -> Result<StopReason<<Self::Arch as Arch>::Usize>, Self::Error>;
+}
+
+define_ext!(SingleThreadReverseContOps, SingleThreadReverseCont);
+
+/// Target Extension - [Reverse stepping] for single threaded targets.
+///
+/// Reverse stepping allows the target to run backwards by one step.
+/// Usually this corresponds to a single instruction.
```
daniel5151, line 134: ditto

daniel5151 (changes requested):
Home stretch!

I've just got a few lingering nits (and `clippy` is getting mad about something in the CI), but aside from that, it's looking good to merge!

Thanks for taking the time to implement and troubleshoot this feature, along with coming along for the ride of iterating on its design 😄

src/target/ext/base/multithread.rs
DrChat: Oops. I could've sworn I removed at least one of these... how many times did I repeat this? 😅

DrChat:
Perfect! Another iteration pushed up to you for review.
Thank you for the help and extremely quick reviews! You made developing this feature a breeze :)

daniel5151 (approved):
Fantastic! Great work!

Now that this has gone in, you've lit a fire under my ass to actually polish up and cut a proper `0.5` release! 😅

I'll try to carve out some time in the next week or two to go that...

daniel5151:
Ah, spoke a bit too soon. looks like you've missed a clippy lint 😉 

I really need to fix my CI to use stable clippy, since I'm fairly sure this lint is only in nightly clippy right now...

DrChat:
Ahh yeah, let me squash my branch and fix that.

daniel5151:
🚀 🚀 🚀

DrChat:
🔥🔥🔥 Appreciate it!

daniel5151:
So, I found some time to pull and test these changes, and I've got a sneaking suspicion that you never tested the `armv4t` example, did you :eyes:

The first issue I found was pretty low-hanging, which was that the `qSupported` response for reverse-continue isn't `ReverseCont+`, but `ReverseContinue+`. I fixed this in a79f666a.

The second issue I found was easy to spot, but required quite a non-trivial fix. Namely, calling `reverse-step` on the single-threaded example would fail due to an issue with how I handled the "any" variant of the `H` packet. While I wouldn't have expected you to find a fix for this issue yourself, I would have at least expected you to _encounter_ this issue during testing, and give the a heads up to fix it.

All that said, it's not a _huge_ that these issues slipped in through this PR, as this code was targeting the dev branch, and I was able to quickly fix the issues myself. Nonetheless, please take this as a learning opportunity, and be mindful to test _all_ the code in your PR before signing off on it.

DrChat:
Oops! Sorry 😛 - I did not test this before pushing it out to you.
At first I didn't think I could test it (before writing the code out for the `armv4t` example) and had that idea stuck in my head.
I was planning on doing more testing when I integrated this code back into our codebase.

Appreciate you catching these issues :)

daniel5151:
In case you were interested, I figured out why the handler code wasn't being DCE'd. It turns out LLVM isn't smart enough to do multi-level IDET elimination... at least not without a bit of prodding.

In a nutshell, it turns out that IDET "toggle" methods need to be annotated with `#[inline(always)]` to guarantee optimization.
Check out this commit for more details: 6bc47ea

DrChat:
Looks like I'm seeing a bug now for some reason - GDB output:
```
Remote register badly formatted: T05replaylog:begin
here:
```

Not sure why. I formatted the replaylog message according to the [GDB documentation](https://sourceware.org/gdb/current/onlinedocs/gdb/Stop-Reply-Packets.html)

PR #71, opened 2021-09-16
Fix crates.io badges

atouchet:
### Description

This PR fixes the crates.io badges in the project Readmes.


daniel5151 (approved):
Ahh, I was wondering why those'd disappeared!

Appreciate the fix!

PR #69, opened 2021-08-22
Support read exec-file

bet4it:
### Description

This PR implements the `qXfer:exec-file:read` command, based off the GDB documentation [here](https://sourceware.org/gdb/current/onlinedocs/gdb/General-Query-Packets.html#qXfer-executable-filename-read).

### API Stability

- [x] This PR does not require a breaking API change

### Checklist

- Implementation
  - [x] `cargo build` compiles without `errors` or `warnings`
  - [x] `cargo clippy` runs without `errors` or `warnings`
  - [x] `cargo fmt` was run
  - [x] All tests pass
- Documentation
  - [x] rustdoc + approprate inline code comments
  - [ ] Updated CHANGELOG.md
  - [x] (if appropriate) Added feature to "Debugging Features" in README.md
- _If implementing a new protocol extension IDET_
  - [x] Included a basic sample implementation in `examples/armv4t`
  - [x] Included output of running `examples/armv4t` with `RUST_LOG=trace` + any relevant GDB output under the "Validation" section below
  - [x] Confirmed that IDET can be optimized away (using `./scripts/test_dead_code_elim.sh` and/or `./example_no_std/check_size.sh`)
  - [ ] **OR** Implementation requires adding non-optional binary bloat (please elaborate under "Description")
- _If upstreaming an `Arch` implementation_
  - [ ] I have tested this code in my project, and to the best of my knowledge, it is working as intended.

### Validation

<details>
<summary>GDB output</summary>

```
Remote debugging using :9001
Reading /test.elf from remote target...
warning: File transfers from remote targets can be slow. Use "set sysroot" to access files locally instead.
Reading /test.elf from remote target...
Reading symbols from target:/test.elf...
main () at test.c:1
1       int main() {
```

</details>

<details>
<summary>armv4t output</summary>

```
    Finished dev [unoptimized + debuginfo] target(s) in 0.02s
     Running `target/debug/examples/armv4t`
loading section ".text" into memory from [0x55550000..0x55550078]
Setting PC to 0x55550000
Waiting for a GDB connection on "127.0.0.1:9001"...
Debugger connected from 127.0.0.1:37052
 TRACE gdbstub::protocol::recv_packet > <-- +
 TRACE gdbstub::protocol::recv_packet > <-- $qSupported:multiprocess+;swbreak+;hwbreak+;qRelocInsn+;fork-events+;vfork-events+;exec-events+;vContSupported+;QThreadEvents+;no-resumed+;xmlRegisters=i386#6a
[...]

Issue #70, opened 2021-09-08
Internal refactor to enable using 100% of PacketBuf as response buffer


daniel5151:
As discussed in #69, we'll gradually want to move various callback based APIs over to a "here is a `&mut [u8]` buffer, please fill it with data" APIs.

Instead of allocating a whole new outgoing packet buffer for this, it'd be nice to reuse the existing `PacketBuf`. This could work, since packets that request data are almost always able to be parsed into fixed-size, lifetime-free structs, which would leave the packet buf available to be used as scratch space.

Unfortunately, the current implementation of the packet parser includes a "late decode" step, whereby target-dependent fields (such as memory addresses / breakpoint kinds / etc...) are parsed into `&[u8]` bufs in the packet parsing code, and are only converted into their concrete types later on, in the handler code (where the type of `Target` is known). This is an important property to maintain, as eventually, we'd want to support debugging multiple `Target` types at the same time (e.g: on macOS, a single gdbserver can debug both x86 and ARM code), and the only way to do this would be by having the packet parsing code be `Target` agnostic.

Instead, there should be some way of obtaining a reference to the entire, raw, underlying `&mut [u8]` `PacketBuf` after the late decode step has been completed, but this is harder than it seems. Getting the lifetimes to line up here will probably be tricky, and I suspect getting this working will require some real code-cotorsion, and possibly even a sprinkle of `unsafe`.

* * *

In the meantime, we'll be going with the approach used by the `m` packet, whereby the packet parsing code will "stash" a `&mut [u8]` pointing to the trailing unused bit of the buffer as part of the parsed struct.

This works, but is a bit wasteful (as not 100% of the packet buffer is being utilized), and also a bit annoying to implement on a per-packet basis. Nonetheless, the GDB RSP allows targets to return less than the requested amount of data provided without there necessarily being an error, specifically because certain implementation might be using different sized buffers for incoming / outgoing data.

Given that "workaround" works pretty well, and that losing ~30 bytes of a ~4096 byte PacketBuf isn't particularly noticeable, getting to 100% efficiency isn't a _super_ high priority, but it's still something to think about, and potentially implement at some point.

bet4it:
We can store the `Range` of `PacketBuf` in `ParseCommand` rather than `&[u8]`?

daniel5151:
Some relevant discussion regarding this issue: https://github.com/daniel5151/gdbstub/pull/72#discussion_r711225603

bet4it:
[...]

PR #72, opened 2021-09-17
Pass PacketBuf as an argument of API

bet4it:
### Description
- Update the other qXfer-backed protocol extensions to use the `(length: usize, offset: u64, buf: &mut [u8])` API
  - At the moment, I think this is just `MemoryMap` and `TargetXmlOverride`
- Update the vFile API to use the `&mut [u8]` PacketBuf instead of the current callback-based approach.
- Update the vFile API to use plain 'ol `usize` for `length` and `offset` parameters, instead of `Target::Arch::Usize`
  - Context: https://github.com/daniel5151/gdbstub/pull/69#discussion_r707477573

_Originally posted by @daniel5151 in https://github.com/daniel5151/gdbstub/issues/69#issuecomment-920159984_

### API Stability

### Checklist

- Implementation
  - [x] `cargo fmt` was run
  - [x] All tests pass
- Documentation
  - [ ] Updated CHANGELOG.md

### DIFF
--- a/examples/armv4t/gdb/host_io.rs
+++ b/examples/armv4t/gdb/host_io.rs
@@ -5,6 +5,7 @@ use gdbstub::target::ext::host_io::{
     FsKind, HostIoErrno, HostIoError, HostIoOpenFlags, HostIoOpenMode, HostIoResult, HostIoStat,
 };
 
+use super::copy_range_to_buf;
 use crate::emu::Emu;
 use crate::TEST_PROGRAM_ELF;
 
@@ -138,12 +139,7 @@ impl target::ext::host_io::HostIoPread for Emu {
     ) -> HostIoResult<usize, Self> {
         if fd < FD_RESERVED {
             if fd == 0 {
-                let len = TEST_PROGRAM_ELF.len();
-                let data =
-                    &TEST_PROGRAM_ELF[len.min(offset as usize)..len.min(offset as usize + count)];
-                let buf = &mut buf[..data.len()];
-                buf.copy_from_slice(data);
-                return Ok(data.len());
+                return Ok(copy_range_to_buf(TEST_PROGRAM_ELF, offset, count, buf));
             } else {
                 return Err(HostIoError::Errno(HostIoErrno::EBADF));
             }
@@ -250,15 +246,11 @@ impl target::ext::host_io::HostIoReadlink for Emu {
         if filename == b"/proc/1/exe" {
             // Support `info proc exe` command
             let exe = b"/test.elf";
-            let buf = &mut buf[..exe.len()];
-            buf.copy_from_slice(exe);
-            return Ok(exe.len());
+            return Ok(copy_range_to_buf(exe, 0, exe.len(), buf));
         } else if filename == b"/proc/1/cwd" {
             // Support `info proc cwd` command
             let cwd = b"/";
-            let buf = &mut buf[..cwd.len()];
-            buf.copy_from_slice(cwd);
-            return Ok(cwd.len());
+            return Ok(copy_range_to_buf(cwd, 0, cwd.len(), buf));
         } else if filename.starts_with(b"/proc") {
             return Err(HostIoError::Errno(HostIoErrno::ENOENT));
         }
@@ -270,10 +262,7 @@ impl target::ext::host_io::HostIoReadlink for Emu {
             .to_str()
             .ok_or(HostIoError::Errno(HostIoErrno::ENOENT))?
             .as_bytes();
-        let len = data.len();
-        let buf = &mut buf[..len];
-        buf.copy_from_slice(data);
-        Ok(len)
+        Ok(copy_range_to_buf(data, 0, data.len(), buf))
     }
 }
 
--- a/examples/armv4t/gdb/host_io.rs
+++ b/examples/armv4t/gdb/host_io.rs
@@ -262,7 +262,11 @@ impl target::ext::host_io::HostIoReadlink for Emu {
             .to_str()
             .ok_or(HostIoError::Errno(HostIoErrno::ENOENT))?
             .as_bytes();
-        Ok(copy_range_to_buf(data, 0, data.len(), buf))
+        if data.len() <= buf.len() {
+            Ok(copy_range_to_buf(data, 0, data.len(), buf))
+        } else {
+            Err(HostIoError::Errno(HostIoErrno::ENAMETOOLONG))
+        }
     }
 }
 
--- a/examples/armv4t/gdb/host_io.rs
+++ b/examples/armv4t/gdb/host_io.rs
@@ -2,8 +2,7 @@ use std::io::{Read, Seek, Write};
 
 use gdbstub::target;
 use gdbstub::target::ext::host_io::{
-    FsKind, HostIoErrno, HostIoError, HostIoOpenFlags, HostIoOpenMode, HostIoOutput, HostIoResult,
-    HostIoStat, HostIoToken,
+    FsKind, HostIoErrno, HostIoError, HostIoOpenFlags, HostIoOpenMode, HostIoResult, HostIoStat,
 };
 
 use crate::emu::Emu;
@@ -133,16 +132,18 @@ impl target::ext::host_io::HostIoPread for Emu {
     fn pread<'a>(
         &mut self,
         fd: u32,
-        count: u32,
-        offset: u32,
-        output: HostIoOutput<'a>,
-    ) -> HostIoResult<HostIoToken<'a>, Self> {
+        count: usize,
+        offset: u64,
+        buf: &mut [u8],
+    ) -> HostIoResult<usize, Self> {
         if fd < FD_RESERVED {
             if fd == 0 {
                 let len = TEST_PROGRAM_ELF.len();
-                return Ok(output.write(
-                    &TEST_PROGRAM_ELF[len.min(offset as usize)..len.min((offset + count) as usize)],
-                ));
+                let data =
+                    &TEST_PROGRAM_ELF[len.min(offset as usize)..len.min(offset as usize + count)];
+                let buf = &mut buf[..data.len()];
+                buf.copy_from_slice(data);
+                return Ok(data.len());
             } else {
                 return Err(HostIoError::Errno(HostIoErrno::EBADF));
             }
@@ -153,10 +154,9 @@ impl target::ext::host_io::HostIoPread for Emu {
             _ => return Err(HostIoError::Errno(HostIoErrno::EBADF)),
         };
 
-        let mut buffer = vec![0; count as usize];
-        file.seek(std::io::SeekFrom::Start(offset as u64))?;
-        let n = file.read(&mut buffer)?;
-        Ok(output.write(&buffer[..n]))
+        file.seek(std::io::SeekFrom::Start(offset))?;
+        let n = file.read(buf)?;
+        Ok(n)
     }
 }
 
@@ -246,29 +246,34 @@ impl target::ext::host_io::HostIoUnlink for Emu {
 }
 
 impl target::ext::host_io::HostIoReadlink for Emu {
-    fn readlink<'a>(
-        &mut self,
-        filename: &[u8],
-        output: HostIoOutput<'a>,
-    ) -> HostIoResult<HostIoToken<'a>, Self> {
+    fn readlink<'a>(&mut self, filename: &[u8], buf: &mut [u8]) -> HostIoResult<usize, Self> {
         if filename == b"/proc/1/exe" {
             // Support `info proc exe` command
-            return Ok(output.write(b"/test.elf"));
+            let exe = b"/test.elf";
+            let buf = &mut buf[..exe.len()];
+            buf.copy_from_slice(exe);
+            return Ok(exe.len());
         } else if filename == b"/proc/1/cwd" {
             // Support `info proc cwd` command
-            return Ok(output.write(b"/"));
+            let cwd = b"/";
+            let buf = &mut buf[..cwd.len()];
+            buf.copy_from_slice(cwd);
+            return Ok(cwd.len());
         } else if filename.starts_with(b"/proc") {
             return Err(HostIoError::Errno(HostIoErrno::ENOENT));
         }
 
         let path =
             std::str::from_utf8(filename).map_err(|_| HostIoError::Errno(HostIoErrno::ENOENT))?;
-        Ok(output.write(
-            std::fs::read_link(path)?
-                .to_str()
-                .ok_or(HostIoError::Errno(HostIoErrno::ENOENT))?
-                .as_bytes(),
-        ))
+        let link = std::fs::read_link(path)?;
+        let data = link
+            .to_str()
+            .ok_or(HostIoError::Errno(HostIoErrno::ENOENT))?
+            .as_bytes();
+        let len = data.len();
+        let buf = &mut buf[..len];
+        buf.copy_from_slice(data);
+        Ok(len)
     }
 }
 
--- a/examples/armv4t/gdb/mod.rs
+++ b/examples/armv4t/gdb/mod.rs
@@ -1,4 +1,4 @@
-use core::convert::TryInto;
+use core::convert::{TryFrom, TryInto};
 
 use armv4t_emu::{reg, Memory};
 use gdbstub::target;
@@ -37,12 +37,16 @@ fn cpu_reg_id(id: ArmCoreRegId) -> Option<u8> {
 }
 
 /// Copy a range of `data` (start at `offset` with a size of `length`) to `buf`.
-/// Return the size of data copied. Return 0 if encounter EOF.
+/// Return the size of data copied. Returns 0 if `offset >= buf.len()`.
 ///
 /// Mainly used by qXfer:_object_:read commands.
 pub fn copy_range_to_buf(data: &[u8], offset: u64, length: usize, buf: &mut [u8]) -> usize {
+    let offset = match usize::try_from(offset) {
+        Ok(v) => v,
+        Err(_) => return 0,
+    };
     let len = data.len();
-    let data = &data[len.min(offset as usize)..len.min(offset as usize + length)];
+    let data = &data[len.min(offset)..len.min(offset + length)];
     let buf = &mut buf[..data.len()];
     buf.copy_from_slice(data);
     data.len()
--- a/examples/armv4t/gdb/mod.rs
+++ b/examples/armv4t/gdb/mod.rs
@@ -36,6 +36,14 @@ fn cpu_reg_id(id: ArmCoreRegId) -> Option<u8> {
     }
 }
 
+pub fn copy_range_to_buf(data: &[u8], offset: u64, length: usize, buf: &mut [u8]) -> usize {
+    let len = data.len();
+    let data = &data[len.min(offset as usize)..len.min(offset as usize + length)];
+    let buf = &mut buf[..data.len()];
+    buf.copy_from_slice(data);
+    data.len()
+}
+
 impl Target for Emu {
     // As an example, I've defined a custom architecture based off
     // `gdbstub_arch::arm::Armv4t`. The implementation is in the `custom_arch`
--- a/examples/armv4t/gdb/mod.rs
+++ b/examples/armv4t/gdb/mod.rs
@@ -36,6 +36,10 @@ fn cpu_reg_id(id: ArmCoreRegId) -> Option<u8> {
     }
 }
 
+/// Copy a range of `data` (start at `offset` with a size of `length`) to `buf`.
+/// Return the size of data copied. Return 0 if encounter EOF.
+///
+/// Mainly used by qXfer:_object_:read commands.
 pub fn copy_range_to_buf(data: &[u8], offset: u64, length: usize, buf: &mut [u8]) -> usize {
     let len = data.len();
     let data = &data[len.min(offset as usize)..len.min(offset as usize + length)];
--- a/src/gdbstub_impl/ext/base.rs
+++ b/src/gdbstub_impl/ext/base.rs
@@ -127,20 +127,29 @@ impl<T: Target, C: Connection> GdbStubImpl<T, C> {
             }
             Base::qXferFeaturesRead(cmd) => {
                 #[allow(clippy::redundant_closure)]
-                let xml = target
-                    .target_description_xml_override()
-                    .map(|ops| ops.target_description_xml())
-                    .or_else(|| T::Arch::target_description_xml());
-
-                match xml {
-                    Some(xml) => {
+                let ret = if let Some(ops) = target.target_description_xml_override() {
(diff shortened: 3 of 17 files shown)


src/protocol/commands/_vFile_readlink.rs
```diff
@@ -3,19 +3,23 @@ use super::prelude::*;
...
     pub filename: &'a [u8],
+
+    pub buf: &'a mut [u8],
 }
 
 impl<'a> ParseCommand<'a> for vFileReadlink<'a> {
     fn from_packet(buf: PacketBuf<'a>) -> Option<Self> {
-        let body = buf.into_body();
+        let (buf, body_range) = buf.into_raw_buf();
+        let (body, buf) = buf[body_range.start..].split_at_mut(body_range.end - body_range.start);
```
daniel5151, line 13: Please refrain from using direct indexing operations from inside `gdbstub`, as it will introduce panicking code into the library. Unfortunately, AFAIK, there isn't a great way to automatically lint for this, so we must remain vigilant.

In this case, if I were writing this code, I'd probably end up using something like `cargo asm` to check the generated asm from this file, and/or throwing the code into godbolt.org and seeing what comes out.

That said, this is pretty tricky to do, and given that this code is in the packet parser for the `vFile` extension, it's not _super_ likely that it'll be used in a project with strict no panic properties...

TL;DR: Please add a comment to this code: `// TODO: rewrite to avoid panic`, and I'll get around to it myself at some point.

src/target/ext/host_io.rs
```diff
@@ -362,11 +338,7 @@ pub trait HostIoReadlink: HostIo {
     /// will consume the `output` object and return a [`HostIoToken`]. This
     /// token ensures that the implementer of this method calls
     /// [`HostIoOutput::write`].
-    fn readlink<'a>(
-        &mut self,
-        filename: &[u8],
-        output: HostIoOutput<'a>,
-    ) -> HostIoResult<HostIoToken<'a>, Self>;
+    fn readlink<'a>(&mut self, filename: &[u8], buf: &mut [u8]) -> HostIoResult<usize, Self>;
```
daniel5151, line 341: interesting... it seems like in this case, it would actually be _impossible_ to re-use the entire `PacketBuf` to send the response. This implies that #70 might not be possible for all cases.

daniel5151 (changes requested):
Nice, looking good. Definitely simplified a lot of the callback-code. Thanks again for pointing out that relevant line in the spec - I'm glad we can simplify this API!

Left a couple of inline comments, but on the whole, this is looking good.

bet4it:
Do I need to change the type of all the `target_description_xml`?

And I really don't like the `copy_from_slice` code all over the place. We may need some awesome macro or function to wrap it?

src/target/ext/host_io.rs
bet4it: It may be not impossible with the current code structure, but we can always find a point where the data in `buf` is used and we don't start to use `buf` to prepare the response, so it's still possible?

If we are allowed to use multiple mutable reference ( like C ), we can assure it will safe to do it?

daniel5151:
> Do I need to change the type of all the `target_description_xml`?

Are you referring to the `target_description_xml` that's part of the `Arch` trait? If so, no, you don't, as that is a piece of static metadata associated with the `Arch`, and will be guaranteed to exist as a slice somewhere in the binary's static memory.

> And I really don't like the `copy_from_slice` code all over the place. We may need some awesome macro or function to wrap it?

From what I gather, most instances of repetitive `copy_from_slice` code occurs in the `armv4t` example code, where you're reading from static buffers? If so, I think it's reasonable to add a helper function somewhere in the crate (e.g: in `examples/armv4t/gdb/mod.rs`) to wrap this common pattern.

Notably, I don't think we want this helper function to be part of the `gdbstub` crate API itself.

src/target/ext/host_io.rs
daniel5151: I'm not entirely sure what you mean... but the point I was trying to make is that so long as the API signature has a `filename: &[u8]` that points somewhere into the middle of the PacketBuf, `buf: &mut [u8]` cannot point to the start of the packet buf, as you would have an aliased mutable pointer, which would be _very bad_ in Rust.

Off the top of my head, if we _really_ wanted to enable this use-case, we could do something like:

```rust
struct ReadLinkPacket<'a> {
    filename_start: usize,
    filename_end: usize,
    buf: &'a mut [u8]
}

impl<'a> ReadLinkPacket<'a> {
    fn filename(&self) -> &[u8] {
        &self.buf[self.filename_start..self.filename_end]
    }

    fn into_buf(self) -> &'a mut [u8] {
        self.buf
    }
}

fn readlink(&mut self, filename_and_buf: ReadLinkPacket<'_>) -> HostIoResult<usize, Self> {
    let filename = filename_and_buf.filename();
    let link = std::fs::read_link(filename) // etc...
    let buf = filename_and_buf.into_buf();
    // compose response using buf
}
```

Also, while thinking about this, I stumbled across an interesting observation: if you check the man page for the underlying `readlink` syscall, you'll find that the filename buffer and the output buffer are actually marked `restrict`, which means it wouldn't be possible to read the resolved link path directly into the buffer containing the filename. i.e: you _must_ either allocate a new buffer to read the resolved path into, or use memory disjoint from the `filename` buf, regardless if you were using Rust or C.

FWIW, I think the current implementation that doesn't use 100% of the PacketBuf is totally reasonable, especially since the underlying packet doesn't even have a mechanism of relaying "partial reads", and assumes that the packetbuf has enough space to fit the response.

bet4it: Let's take a look at how GDB implements it:
```C
#if !defined (PATH_MAX) || (PATH_MAX > (PBUFSIZ / 2 + 1))
#  define HOSTIO_PATH_MAX (PBUFSIZ / 2 + 1)
#else
#  define HOSTIO_PATH_MAX PATH_MAX
#endif
```
```C
static void
handle_readlink (char *own_buf, int *new_packet_len)
{
  char filename[HOSTIO_PATH_MAX], linkname[HOSTIO_PATH_MAX];
  char *p;
  int ret, bytes_sent;

  p = own_buf + strlen ("vFile:readlink:");

  if (require_filename (&p, filename)
      || require_end (p))
    {
      hostio_packet_error (own_buf);
      return;
    }

  if (hostio_fs_pid != 0 && the_target->multifs_readlink != NULL)
    ret = the_target->multifs_readlink (hostio_fs_pid, filename,
					linkname,
					sizeof (linkname) - 1);
  else
    ret = readlink (filename, linkname, sizeof (linkname) - 1);

  if (ret == -1)
    {
      hostio_error (own_buf);
      return;
    }

  bytes_sent = hostio_reply_with_data (own_buf, linkname, ret, new_packet_len);

  /* If the response does not fit into a single packet, do not attempt
     to return a partial response, but simply fail.  */
  if (bytes_sent < ret)
    sprintf (own_buf, "F-1,%x", FILEIO_ENAMETOOLONG);
}
```
So we have two ways to implement it:
1. Copy the filename into a temporary buf allocated on the stack and pass this buf to `ops.readlink`.
2. Allocate two global buf with the same size: `PacketBufIn` and `PacketBufOut`. This will cost twice of the memory, but the implementation will be very easy.

daniel5151: > 1. Copy the filename into a temporary buf allocated on the stack and pass this buf to `ops.readlink`.

This forces `gdbstub` to arbitrarily decide what a reasonable "max path" size would be, which is something I've avoided. This is actually one of those cases where a callback-based output function would be ideal, since it gives the calling code the ability to allocate a stack-allocated buffer of whatever size it deems appropriate, and pass it back into `gdbstub` code.



> 2. Allocate two global buf with the same size: `PacketBufIn` and `PacketBufOut`. This will cost twice of the memory, but the implementation will be very easy.

I feel like this is something I brought up in the past (or maybe I just thought of it and didn't write it down), but yet, this would be the easiest way to handle these sorts of cases. The `gdbstub` build could be modified to accept an optional `PacketBufOut`, and the handler code would check for it's existence when deciding which `PacketBuf` to pass to the target implementation.

examples/armv4t/gdb/mod.rs
```diff
@@ -36,6 +36,14 @@ fn cpu_reg_id(id: ArmCoreRegId) -> Option<u8> {
     }
 }
 
+pub fn copy_range_to_buf(data: &[u8], offset: u64, length: usize, buf: &mut [u8]) -> usize {
```
daniel5151, line 39: please add a (brief) comment documenting the purpose and behavior of this function (notably, its behavior with EOF conditions)

src/gdbstub_impl/ext/base.rs
```diff
@@ -126,21 +126,31 @@ impl<T: Target, C: Connection> GdbStubImpl<T, C> {
...
+                    buf.copy_from_slice(data);
+                    data.len()
+                } else {
                     // If the target hasn't provided their own XML, then the initial response to
-                    // "qSupported" wouldn't have included "qXfer:features:read", and gdb wouldn't
-                    // send this packet unless it was explicitly marked as supported.
-                    None => return Err(Error::PacketUnexpected),
+                    // "qSupported" wouldn't have included "qXfer:features:read", and gdb
+                    // wouldn't send this packet unless it was
+                    // explicitly marked as supported.
```
daniel5151, line 144: nit: fix the text wrapping to match the original comment

daniel5151 (changes requested):
a few nits, but I think this is about ready to merge (feel free to un-draft)

daniel5151 (changes requested):

src/target/ext/host_io.rs
bet4it: > This forces `gdbstub` to arbitrarily decide what a reasonable "max path" size would be, which is something I've avoided. This is actually one of those cases where a callback-based output function would be ideal, since it gives the calling code the ability to allocate a stack-allocated buffer of whatever size it deems appropriate, and pass it back into `gdbstub` code.

What I said don't need `gdbstub` to decide "max path" size, because we already know the real length of `cmd.filename`, and allocate the buffer as needed.

> I feel like this is something I brought up in the past (or maybe I just thought of it and didn't write it down), but yet, this would be the easiest way to handle these sorts of cases. The `gdbstub` build could be modified to accept an optional `PacketBufOut`, and the handler code would check for it's existence when deciding which `PacketBuf` to pass to the target implementation.

It will be a mess. We pass entire `PacketBuf` to avoid possibility of overflow, if you don't know if the buffer will lead to overflow, how to write the code...
I think we can just split the `PacketBuf` user provided into two, and set `PacketSize` to half of the length of `PacketBuf`.

bet4it:
We don't pass the entire `PacketBuf` to `readlink` now, do we need to consider the situation that `buf` is not enough to hold the result of `readlink`?

src/target/ext/host_io.rs
daniel5151: > What I said don't need `gdbstub` to decide "max path" size, because we already know the real length of `cmd.filename`, and allocate the buffer as needed.

Maybe I misunderstood what you meant... could you sketch out a rough version of the code you're envisioning?

> I think we can just split the `PacketBuf` user provided into two, and set `PacketSize` to half of the length of `PacketBuf`.

The issue with this approach (or my approach for that matter), its that almost every other packet doesn't have this limitation, which means that you'd just end allocating a bunch of memory you're not using all that often.

Admittedly, "a bunch of memory" isn't really appropriate when we're talking about a 4k buffer on a PC with gigabytes of RAM, but if you consider `gdbstub` running on a embedded platform, we can't just waste memory like that willy-nilly (hence my reasoning for making it an explicitly optional bit of functionality, which individual implementations can opt into as part of the `gdbstub` constructor). Obviously, it's more messy, but that's the price you pay for modularity ¯\\\_(ツ)_/¯

daniel5151:
> We don't pass the entire `PacketBuf` to `readlink` now, do we need to consider the situation that `buf` is not enough to hold the result of `readlink`?

You've seen the reference `gdbserver` code yourself:

```c
  /* If the response does not fit into a single packet, do not attempt
     to return a partial response, but simply fail.  */
  if (bytes_sent < ret)
    sprintf (own_buf, "F-1,%x", FILEIO_ENAMETOOLONG);
```

i.e: this detail is something the user should handle themselves.

...that said, I think this is an important point worth stressing in the `readlink` docs. I'll leave a suggested doc change as a review comment...

src/target/ext/host_io.rs
```diff
@@ -358,15 +335,8 @@ define_ext!(HostIoUnlinkOps, HostIoUnlink);
...
-    /// The data read _must_ be sent by calling [`HostIoOutput::write`], which
-    /// will consume the `output` object and return a [`HostIoToken`]. This
-    /// token ensures that the implementer of this method calls
-    /// [`HostIoOutput::write`].
-    fn readlink<'a>(
-        &mut self,
-        filename: &[u8],
-        output: HostIoOutput<'a>,
-    ) -> HostIoResult<HostIoToken<'a>, Self>;
+    /// Return the number of bytes written into `buf`.
```
daniel5151, line 338: update the docs with a note regarding what to do if the buffer doesn't have enough space.

```suggestion
    /// Return the number of bytes written into `buf`.
    ///
    /// Unlike most other Host IO handlers, if the resolved file path exceeds
    /// the length of the provided `buf`, the target should NOT return a
    /// partial response, and MUST return a `Err(HostIoErrno::ENAMETOOLONG)`.
```

daniel5151 (changes requested):

src/target/ext/host_io.rs
bet4it: It's just:
```rust
HostIo::vFileReadlink(cmd) if ops.enable_readlink().is_some() => {
    let ops = ops.enable_readlink().unwrap();
    let mut stack_buf = [0; cmd.filename.len()];
    stack_buf.copy_from_slice(cmd.filename)
    handle_hostio_result! {
        if let Ok(ret) = ops.readlink(stack_buf, cmd.buf) => {
            res.write_str("F")?;
            res.write_num(ret)?;
            res.write_str(";")?;
            res.write_binary(cmd.buf.get(..ret).ok_or(Error::PacketBufferOverflow)?)?;
        }
    };
    HandlerStatus::Handled
}
```
Or if we choose `ReadLinkPacket` to solve multiple mutable reference problem:
```rust
let filename = filename_and_buf.filename();
let mut stack_buf = [0; filename.len()];
stack_buf.copy_from_slice(cmd.filename)
let buf = filename_and_buf.into_buf();
handle_hostio_result! {
    if let Ok(ret) = ops.readlink(stack_buf, buf) => {
        ......
    }
};
```
`ReadLinkPacket` is too ugly that it's impossible for us to use it as an argument of API, but we can use it in `gdbstub` internal code.

bet4it:
The reference is about the situation that the length of `readlink` result is larger than `PacketSize`.
But now the length of `buf` argument may be much smaller than `PacketSize`. For example, `PacketSize` is set to 500, and `filename` is 490, there may no extra spaces in `buf`, so `readlink` will always fail, but the situation shouldn't fail in actually.

src/target/ext/host_io.rs
daniel5151: Ahh, I see.

Unfortunately, `[0; cmd.filename.len()]` isn't valid Rust, since Rust doesn't support variable-length stack-allocated arrays (i.e: `alloca`). I'll admit, this is one of those features that I do hope Rust gains at some point, as while it's certainly dangerous and prone to misuse, it's a powerful tool for this kind of resource constrained systems programming...

This ties back to my earlier comment about having to decide on an "arbitrary max length", since the length of this stack allocated buffer would have to be a constant value. With sufficient plumbing, we could make this a bit "smarter" by making the `gdbstub` constructor take a const-generic value for the PacketBuf len, and then use that value as a stack allocated buf, but that also seems like a bad idea (i.e: `gdbstub` can be driven from IRQ contexts, with very limited stack space).

To make a long story short: it seems that the "solution" to using 100% of the PacketBuf in these sorts of cases will be to either do some serious API contortions (e.g: using that funky `ReadLinkPacket` code I sketched out earlier), or to throw our hands in the air, and require that users provide their own outgoing packet buffer for these sorts of cases.

daniel5151:
I see... this ties back into the discussion we are having here: https://github.com/daniel5151/gdbstub/pull/72#discussion_r711225603

Let's keep the conversation focused, and continue discussing this on that particular comment thread.

src/target/ext/host_io.rs
daniel5151: I think the key meta-insight with this particular packet is that it seems to require that _someone_ allocate an extra buffer _at some point_, either to store a copy of the provided `filename` (as you're suggesting), or to compose a response (as I'm suggesting). 

The reference implementation went with the first approach, where they picked an "arbitrary" max filename (based on either a user-defined `define`, or half the `PacketBuf` len) and then allocated it themselves. I'm not sure if I'm a huge fan of this approach from a "philosophically elegant code" POV, but it certainly seems like a reasonable "engineering realities" solution to the problem.

In `gdbstub`, I try my hardest to avoid these sorts of "arbitrary internal constants", and offer users the choice of how they'd like to handle these sorts of cases - either via a more complex API (e.g: using callbacks, type-shenanigans, etc...), or via internal switch statements (i.e: "give me another buffer as part of the constructor, and I'll use that one instead of the remaining bit of the packetbuf").

IMO, I think the current implementation is perfectly reasonable. We don't specify an arbitrary max file path, and if a future user runs into this limitation, I'd imagine one of two things will happen:

1. they'll just bump PacketSize up to compensate
2. They'll file an issue, at which point we can review their particular limitations, and decide on an approach that works for them

bet4it: > we could make this a bit "smarter" by making the `gdbstub` constructor take a const-generic value for the PacketBuf len, and then use that value as a stack allocated buf, but that also seems like a bad idea (i.e: `gdbstub` can be driven from IRQ contexts, with very limited stack space).

In such situation the user won't implement Host I/O, so we don't need to consider it?

> either do some serious API contortions (e.g: using that funky `ReadLinkPacket` code I sketched out earlier)

Didn't you say that only `ReadLinkPacket` is not enough, we still need a extra buffer?

bet4it:
What I want to say here is that if user don't provide `PacketBufOut`, the behavior of `readlink` is not suitable in some situations.

src/target/ext/host_io.rs
daniel5151: Yes, precisely: the meta argument I'm trying to make here is that it might not _matter_ that there are certain degenerate cases where this "partial buffer" approach will fail, since those cases are only likely to arise when running on a resource constrained platform where Host IO likely isn't being implemented.

> Didn't you say that only `ReadLinkPacket` is not enough, we still need a extra buffer?

Indeed, except this extra buffer would be allocated by the implementation itself (i.e: as part of the Target's handler implementation), as opposed to internally within `gdbstub`.

daniel5151:
Yep, I totally get that. The question now becomes whether or not those situations are "reasonable", and whether this particular PR should try to address these situations.

IMO, I think the answer to both of those questions is "no", and that we ought to merge this PR now. If this is an edge case you're particularly worried about, we can open a tracking issue (summarizing our discussion in https://github.com/daniel5151/gdbstub/pull/72#discussion_r711225603) about this limitation, and reconsider how/whether this should be fixed at some later point (e.g: once a real-world example of the issue crops up, and/or at some point before committing to a hypothetical stable 1.0.0 API)

src/target/ext/host_io.rs
bet4it: > With sufficient plumbing, we could make this a bit "smarter" by making the `gdbstub` constructor take a const-generic value for the PacketBuf len, and then use that value as a stack allocated buf

Then I think it's not a bad idea? We can implement this way.

daniel5151: As I mentioned in https://github.com/daniel5151/gdbstub/pull/72#issuecomment-922603362, I would like to shelve this discussion for now, and if you feel strongly about this limitation, feel free to open a new tracking issue for it, linking back to this thread + summarizing the various options both you and I have proposed.

daniel5151: i.e: please give this comment a thumbs up if you're good to merge the PR as is, and I'll give it one final review to approve the code, at which point we'll merge this into `dev/0.6`.

bet4it: I think we can continue this discussion in #70?

daniel5151: Ahh, yes, I am a silly man who is reviewing code far too late at night. Yes, we can certainly just continue discussing this at #70

examples/armv4t/gdb/host_io.rs
```diff
@@ -246,29 +242,27 @@ impl target::ext::host_io::HostIoUnlink for Emu {
 }
 
 impl target::ext::host_io::HostIoReadlink for Emu {
-    fn readlink<'a>(
-        &mut self,
-        filename: &[u8],
-        output: HostIoOutput<'a>,
-    ) -> HostIoResult<HostIoToken<'a>, Self> {
+    fn readlink<'a>(&mut self, filename: &[u8], buf: &mut [u8]) -> HostIoResult<usize, Self> {
```
daniel5151, line 245: ahh shoot. based on our discussion, this current implementation is incorrect.

naively using `copy_range_to_buf` is incorrect, as it will silently truncate the response without reporting `ENAMETOOLONG`.

Instead, I think you need to do something like:

```rust
let n = copy_range_to_buf(src, ...);
if n != src.len() {
    return Err(HostIoError::Errno(HostIoErrno::ENAMETOOLONG))
}
```

if you don't like how verbose this is, you could wrap this in a `host_io.rs` local helper function.

daniel5151 (changes requested):
a few last minute things I noticed...

examples/armv4t/gdb/host_io.rs
```diff
@@ -262,7 +262,11 @@ impl target::ext::host_io::HostIoReadlink for Emu {
             .to_str()
             .ok_or(HostIoError::Errno(HostIoErrno::ENOENT))?
             .as_bytes();
-        Ok(copy_range_to_buf(data, 0, data.len(), buf))
+        if data.len() <= buf.len() {
```
daniel5151, line 265: this is fine... but you didn't add the this logic to the hard-coded file names above.

yes, I _know_ that it's incredibly unlikely that the buffer wouldn't be able to contain those strings, but the point of example code is to show off what a reasonably correct implementation looks like.

Please update the `copy_range_to_buf` blocks above with the same guards.

daniel5151 (changes requested):

examples/armv4t/gdb/host_io.rs
bet4it: I didn't add it because RSP packet has the minimum size:
>  If this stub feature is not supported, GDB guesses based on the size of the ‘g’ packet response.

daniel5151: bahhh, you're probably right. on second thought, this seems fiiiiiiine

daniel5151 (approved):
This LGTM! 

I'll go ahead and merge it (unless you have any last-minute objections)

PR #75, opened 2021-09-23
Update dependencies

bet4it:


daniel5151:
Looks like you need to apply some similar fixes to `examples/armv4t_multicore/emu.rs`.

Aside from that, LGTM 👍

bet4it:
It seems that everything works fine

PR #78, opened 2021-09-24
Enable logging for responses with only alloc.

gz:
### Description

Enable logging for responses (as discussed here https://github.com/daniel5151/gdbstub/discussions/77)

### API Stability

### Checklist

- Implementation
  - [x] `cargo fmt` was run
  - [x] All tests pass

### DIFF
--- a/src/protocol/response_writer.rs
+++ b/src/protocol/response_writer.rs
@@ -1,3 +1,6 @@
+use alloc::vec::Vec;
+use alloc::string::String;
+
 use num_traits::PrimInt;
 
 use crate::internal::BeBytes;
@@ -24,7 +27,7 @@ pub struct ResponseWriter<'a, C: Connection + 'a> {
     rle_char: u8,
     rle_repeat: u8,
     // buffer to log outgoing packets. only allocates if logging is enabled.
-    #[cfg(feature = "std")]
+    #[cfg(feature = "alloc")]
     msg: Vec<u8>,
 }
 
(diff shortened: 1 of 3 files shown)


src/protocol/response_writer.rs
```diff
@@ -1,3 +1,6 @@
+use alloc::vec::Vec;
+use alloc::string::String;
```
daniel5151, line 2: These imports will need to be `cfg` gated, as this will result in a compilation error in non `alloc` builds.

daniel5151 (changes requested):
Wow, just as I wrap up opening #79, I realize you've opened a PR fixing it. Thank you!

As per my suggestion in #79, instead of overloading the `"alloc"` feature here, can you add a new feature to `Cargo.toml` called `"trace-pkt"`? This feature would have a dependency on `alloc`, while not forcing this additional logging overhead on all `alloc` users.

You'd also want to cfg-gate the incoming packet trace statement, which is currently always enabled.

daniel5151:
Ah, also, I'd make `"trace-pkt"` a on-by-default feature as well.

src/protocol/response_writer.rs
gz: Fixed

daniel5151 (changes requested):
> You'd also want to cfg-gate the incoming packet trace statement, which is currently always enabled.

i.e: please also `feature = "trace-pkt"` gate the following lines:

https://github.com/daniel5151/gdbstub/blob/dev/0.6/src/protocol/recv_packet.rs#L56-L59
https://github.com/daniel5151/gdbstub/blob/dev/0.6/src/protocol/recv_packet.rs#L114-L117

gz:
> You'd also want to cfg-gate the incoming packet trace statement, which is currently always enabled.

I think I did this in 9e060d2 but let me know if I missed something.

daniel5151 (approved):

daniel5151:
Awesome, thanks again!

PR #84, opened 2021-09-25
Make sure we have 16 Xmm register ids.

### LINKED ISSUE #83: Missing register ids for xmm registers?
I'm getting a `write_register` request for reg_id (57) that isn't really known by gdbstub. I suppose I can add it myself? I just can't really tell which one is supposed to be 57 [here](https://github.com/bminor/binutils-gdb/blob/master/gdb/features/i386/64bit-sse.xml)?

I found this by accident because my gdb sends a packet for setting reg 57/0x39, which is ignored by gdbstub (here with some added log statements):
```
28275229934 [TRACE] - gdbstub::protocol::recv_packet: <-- $P39=ffffffffffffffff#59
28279335634 [INFO ] - gdbstub::gdbstub_impl::ext::single_register_access: got P P { reg_id: 57, val: [255, 255, 255, 255, 255, 255, 255, 255] }
28285262982 [INFO ] - gdbstub::gdbstub_impl::ext::single_register_access: reg is None
28289378772 [INFO ] - gdbstub::gdbstub_impl::ext::single_register_access: empty pkt?
28293528816 [TRACE] - gdbstub::protocol::response_writer: --> $#00
```

gz:
Fixes #83.

### DIFF
--- a/gdbstub_arch/src/x86/reg/id.rs
+++ b/gdbstub_arch/src/x86/reg/id.rs
@@ -177,8 +177,8 @@ impl RegId for X86_64CoreRegId {
             18..=23 => (Segment(X86SegmentRegId::from_u8(id as u8 - 18)?), 4),
             24..=31 => (St(id as u8 - 24), 10),
             32..=39 => (Fpu(X87FpuInternalRegId::from_u8(id as u8 - 32)?), 4),
-            40..=55 => (Xmm(id as u8 - 40), 16),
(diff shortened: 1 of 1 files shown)


gz:
- [ ] fix test 
- [ ] figure out if [we need to adjust the size too](https://github.com/daniel5151/gdbstub/issues/83#issuecomment-927093930)

gz:
Disregard that I should go to sleep.

PR #58, opened 2021-05-27
Dynamic `RegId` Support

DrChat:
### Description
As per a discussion in #53, ([comment](https://github.com/daniel5151/gdbstub/issues/53#issuecomment-849766528)), `gdbstub`'s `RegId` abstraction does not have a way to support dynamic register files due to the register size field returned from `RegId::from_raw_id`.

As suggested by @daniel5151, we can make the register size an optional hint, and enforce register sizing based on that.

This changeset also removes the array allocated in the single register access handler, and instead passes the target a closure that behaves in a similar manner to the one passed to `Registers::gdb_serialize`.

### API Stability

This will break SemVer compatibility, though for a good reason.

### Checklist

- Implementation
  - [x] `cargo build` compiles without `errors` or `warnings`
  - [x] `cargo clippy` runs without `errors` or `warnings`
  - [x] `cargo fmt` was run
  - [x] All tests pass
- Documentation
  - [ ] rustdoc + approprate inline code comments
  - [ ] Updated CHANGELOG.md
  - [ ] (if appropriate) Added feature to "Debugging Features" in README.md

### Validation

TODO...
P.S: How do I query a single register using a GDB client?

<details>
<summary>GDB output</summary>

```
TODO...
```

</details>

<details>
<summary>armv4t output</summary>

```
TODO...
```
</details>

PR #85, opened 2021-09-26
Dynamic `read_register` + `RegId` support

### LINKED ISSUE #80: RIP size on x86-64
[rip size](https://github.com/daniel5151/gdbstub/blob/52ae3c7cf2cb7298925a1ba5b4ed1e904ae64416/gdbstub_arch/src/x86/reg/id.rs#L175) is currently 4 bytes, but maybe it should be 8?

https://github.com/bminor/binutils-gdb/blob/master/gdb/features/i386/64bit-core.xml lists it as:

```xml
  <reg name="rip" bitsize="64" type="code_ptr"/>
```

bet4it:
### Description

New approach to #58.

Part of #70.

Closes #58. Closes #80.

### API Stability

### Checklist

- Implementation
  - [x] `cargo fmt` was run
  - [x] All tests pass
- Documentation
  - [ ] Updated CHANGELOG.md

### DIFF
--- a/src/arch.rs
+++ b/src/arch.rs
@@ -15,7 +15,7 @@
 //! > Having community-created `Arch` implementations distributed in a separate
 //! crate helps minimize any unnecessary "version churn" in `gdbstub` core.
 
-use core::fmt::Debug;
+use core::{fmt::Debug, num::NonZeroUsize};
 
 use num_traits::{FromPrimitive, PrimInt, Unsigned};
 
@@ -28,15 +28,20 @@ use crate::internal::{BeBytes, LeBytes};
 ///
 /// [single register accesses]: crate::target::ext::base::SingleRegisterAccess
 pub trait RegId: Sized + Debug {
-    /// Map raw GDB register number corresponding `RegId` and register size.
+    /// Map raw GDB register number to a corresponding `RegId` and optional
+    /// register size.
+    ///
+    /// If the register size is specified here, gdbstub will include a runtime
+    /// check that ensures target implementations do not send back more
+    /// bytes than the register allows.
     ///
     /// Returns `None` if the register is not available.
-    fn from_raw_id(id: usize) -> Option<(Self, usize)>;
+    fn from_raw_id(id: usize) -> Option<(Self, Option<NonZeroUsize>)>;
 }
 
 /// Stub implementation -- Returns `None` for all raw IDs.
 impl RegId for () {
-    fn from_raw_id(_id: usize) -> Option<(Self, usize)> {
+    fn from_raw_id(_id: usize) -> Option<(Self, Option<NonZeroUsize>)> {
         None
     }
 }
--- a/src/arch.rs
+++ b/src/arch.rs
@@ -28,15 +28,15 @@ use crate::internal::{BeBytes, LeBytes};
 ///
 /// [single register accesses]: crate::target::ext::base::SingleRegisterAccess
 pub trait RegId: Sized + Debug {
-    /// Map raw GDB register number corresponding `RegId`.
+    /// Map raw GDB register number corresponding `RegId` and register size.
     ///
     /// Returns `None` if the register is not available.
-    fn from_raw_id(id: usize) -> Option<Self>;
+    fn from_raw_id(id: usize) -> Option<(Self, usize)>;
 }
 
 /// Stub implementation -- Returns `None` for all raw IDs.
 impl RegId for () {
-    fn from_raw_id(_id: usize) -> Option<Self> {
+    fn from_raw_id(_id: usize) -> Option<(Self, usize)> {
         None
     }
 }
--- a/src/arch.rs
+++ b/src/arch.rs
@@ -15,7 +15,8 @@
 //! > Having community-created `Arch` implementations distributed in a separate
 //! crate helps minimize any unnecessary "version churn" in `gdbstub` core.
 
-use core::{fmt::Debug, num::NonZeroUsize};
+use core::fmt::Debug;
+use core::num::NonZeroUsize;
 
 use num_traits::{FromPrimitive, PrimInt, Unsigned};
 
--- a/src/arch.rs
+++ b/src/arch.rs
@@ -28,15 +28,15 @@ use crate::internal::{BeBytes, LeBytes};
 ///
 /// [single register accesses]: crate::target::ext::base::SingleRegisterAccess
 pub trait RegId: Sized + Debug {
-    /// Map raw GDB register number corresponding `RegId` and register size.
+    /// Map raw GDB register number corresponding `RegId`.
     ///
     /// Returns `None` if the register is not available.
-    fn from_raw_id(id: usize) -> Option<(Self, usize)>;
+    fn from_raw_id(id: usize) -> Option<Self>;
 }
 
 /// Stub implementation -- Returns `None` for all raw IDs.
 impl RegId for () {
-    fn from_raw_id(_id: usize) -> Option<(Self, usize)> {
+    fn from_raw_id(_id: usize) -> Option<Self> {
         None
     }
 }
--- a/src/gdbstub_impl/ext/single_register_access.rs
+++ b/src/gdbstub_impl/ext/single_register_access.rs
@@ -19,10 +19,20 @@ impl<T: Target, C: Connection> GdbStubImpl<T, C> {
                     None => return Ok(HandlerStatus::Handled),
                     Some(v) => v,
                 };
-                let buf = &mut p.buf[0..reg_size];
+                let mut buf = p.buf;
+                if let Some(size) = reg_size {
+                    buf = &mut buf[0..size.get()];
+                }
+
                 let len = ops.read_register(id, reg_id, buf).handle_error()?;
 
-                let buf = &mut buf[0..len];
+                if let Some(size) = reg_size {
+                    if size.get() != len {
+                        return Err(Error::TargetMismatch);
+                    }
+                } else {
+                    buf = &mut buf[0..len];
+                }
                 res.write_hex_buf(buf)?;
                 HandlerStatus::Handled
             }
--- a/src/gdbstub_impl/ext/single_register_access.rs
+++ b/src/gdbstub_impl/ext/single_register_access.rs
@@ -13,13 +13,13 @@ impl<T: Target, C: Connection> GdbStubImpl<T, C> {
     ) -> Result<HandlerStatus, Error<T::Error, C::Error>> {
         let handler_status = match command {
             SingleRegisterAccess::p(p) => {
-                let buf = p.buf;
                 let reg = <T::Arch as Arch>::RegId::from_raw_id(p.reg_id);
-                let reg_id = match reg {
+                let (reg_id, reg_size) = match reg {
                     // empty packet indicates unrecognized query
                     None => return Ok(HandlerStatus::Handled),
                     Some(v) => v,
                 };
+                let buf = &mut p.buf[0..reg_size];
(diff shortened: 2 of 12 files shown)


bet4it:
We may also need to do similar changes to the `g` command?

daniel5151:
Hmm, it seems this change is doing a few different things:

- Instead of stack-allocating a buffer for the `read/write_register` command, you're instead proposing we use the packet buffer.
- You want to remove the register size from `from_raw_id`, and force the user to return the number of bytes the register contained.

That first change seems totally reasonable. That second one... well, I'll need a better explanation for that.

The API you're proposing makes it very easy for a target to inadvertently send too much / too little data to the GDB client. The whole point of encoding the expected register size as part of the `RegId` trait is so that `gdbstub` can provide a "guard rail" to implementations, making sure they send back the right amount of data.

Could you elaborate on why you want to remove this guard rail? Am I missing something?

Note that if this is an attempt to revive/complete #58, then I would much prefer the approach taken in that PR, where the "guard-rail" becomes optional vs. removing it entirely. I think there's a lot of merit in making it "hard to do the wrong thing", which has been a guiding principle in my approach to `gdbstub`'s API and design.

daniel5151:
> We may also need to do similar changes to the `g` command?

Could you elaborate on what you mean? The `g` and `G` packets use an entirely different approach, whereby the API doesn't expose the raw input/output buffer to the user. Instead, the `Register` trait performs a translation between the raw byte stream, and a nicely structured `struct` of registers which the API passes to the implementation.

bet4it:
> The whole point of encoding the expected register size as part of the `RegId` trait is so that `gdbstub` can provide a "guard rail" to implementations, making sure they send back the right amount of data.

Fine. I thought you use register size because you don't know how long the buffer should be passed. If you think the size guard is necessary, I will add it back.

I just thought that, if the user want to fill `dst`, he must already know the size needed. You think if he really don't know, he could get a tip from `dst.len()`? Or we want to rely on the panic of `copy_from_slice`?

> The API you're proposing makes it very easy for a target to inadvertently send too much / too little data to the GDB client.

Is this so important? For example, if I want to report a 64-bit register is 0, can't I just reply "00"? ( Not tested )

> Note that if this is an attempt to revive/complete #58, then I would much prefer the approach taken in that PR, where the "guard-rail" becomes optional vs. removing it entirely.

Then how to design the API if we replace `usize` with `Option<NonZeroUsize>`?

> whereby the API doesn't expose the raw input/output buffer to the user

Oh, you remind me.

What I want to do is replacing the `write_byte` callback of `gdb_serialize` to `PacketBuf`, so we can collect all the registers in this buffer and send them out together. (Remember the discussion of `write_all` in #69?)

daniel5151:
> I just thought that, if the user want to fill `dst`, he must already know the size needed. You think if he really don't know, he could get a tip from `dst.len()`? Or we want to rely on the panic of `copy_from_slice`?

Truncating the buffer to the required length makes it possible to use `try_into()` to convert the `&[u8]` into a `&[u8; _]` (i.e: between a reference to a variable length slice vs. a reference to a fixed-length array), which can simplify the implementation.

> Is this so important? For example, if I want to report a 64-bit register is 0, can't I just reply "00"? ( Not tested )

As per the spec: "The register number n is in hexadecimal, and r… contains two hex digits for each byte in the register (target byte order)."

So it seems the response needs to be _exactly_ the right size.

> Then how to design the API if we replace `usize` with `Option<NonZeroUsize>`?

I was thinking it would look something like:

```rust
    fn read_register(
        &mut self,
        tid: Id,
        reg_id: <Self::Arch as Arch>::RegId,
        buf: &mut [u8],
    ) -> TargetResult<usize, Self>;
```

1. If a size-hint was provided: the `PacketBuf` is truncated to the expected length prior to being passed to the handler (for the aforementioned `try_into` convenience), and the implementation simply returns `buf.len()` from the method (i.e: it used the entire buffer). Within `gdbstub`, we double-check that the returned length matches the size-hint, in case a faulty target implementation accidently wrote too little / too much data.
2. If no size-hint is provided (i.e: to enable the "dynamic RegId" use-case) the `PacketBuf` is _not_ truncated, and `gdbstub` trusts that the target wrote the expected number of bytes into `buf`.

To reiterate: the rationale behind including the guard-rails in the API is realizing the following key aspect of implementing a GDB stub: it is _very easy_ to mess up a handler implementation, and then have things Not Work:tm: for very non-obvious reasons. I love it when libraries double-check things for me, and return a useful error when I've messed up my end of the contract, thereby making my life a lot easier. 

That's my philosophy with `gdbtsub`: folks using the library aren't using it because they're keen on reading the GDB RSP, and the details of how the debugging protocol works... they're just trying to get a stub working to debug the code they're _actually_ working on! A such, `gdbstub` tries really hard to make the integration as painless as possible - including foreseeing the inevitable human mistakes we all make 😄 

* * *

As a side-note: at some point, I hope to eventually revisit the idea I linked to in https://github.com/daniel5151/gdbstub/issues/80#issuecomment-927014535, and see if there's some way of tying a fixed-size array / actual native number type with each `RegId` variant. i.e: having variants like `TargetRegId::Some32BitReg(&mut u32)`, which are passed to the `read/write_register` handlers, making it easy to copy data into/out of the enum without resorting to manual `to/from_le_bytes()` shenanigans.

This would require a _lot_ of changes, and I still don't have a clear idea of how to best design / implement this.

Nonetheless, I thought I'd mention this idea to you, for two reasons:

1. I'd be interested to hear your opinions on this theoretical "radical overhaul" of the API
2. To make it clear that I don't mind this "iterative improvement" of the API you're proposing in this PR, and that it would most likely stay as-is for the next little while.

* * *

> What I want to do is replacing the `write_byte` callback of `gdb_serialize` to `PacketBuf`, so we can collect all the registers in this buffer and send them out together. (Remember the discussion of `write_all` in #69?)

Yep, I'm all for this change. It would certainly make those implementations a _lot_ cleaner, so if you're keen on improving that aspect of `gdbstub`'s internal plumbing, I'd be more than happy to review + merge it in :)

bet4it:
> I'd be interested to hear your opinions on this theoretical "radical overhaul" of the API

This reminds me that why we need both `read_register` and `read_registers`? We only need `read_register`, and pass a reference of one field of `Arch::Registers` struct as an argument to store the result, `read_registers` can be done by calling `read_register` with every register id.

> What I want to do is replacing the `write_byte` callback of `gdb_serialize` to `PacketBuf`, so we can collect all the registers in this buffer and send them out together.

If I do this, I need to replace all the `write_byte` in `gdb_serialize` with copy the data to different part of buffer. Is there any easy way to do this? I may need something like this [`put_u8`](https://docs.rs/bytes/1.1.0/bytes/struct.BytesMut.html#method.put_u8).

daniel5151:
> why we need both `read_register` and `read_registers`

As per the spec: "At a minimum, a stub is required to support the ‘?’ command to tell GDB the reason for halting, ‘g’ and ‘G’ commands for register access, and the ‘m’ and ‘M’ commands for memory access."

https://sourceware.org/gdb/current/onlinedocs/gdb/Overview.html

i.e: single register reads/writes are optional, and shouldn't be a required part of a target implementation.

...though now that I re-read your question, I think you're actually suggesting something else: "why not implement the `g`/`G` packets via the `read/write_register` methods in the user-level API, removing the current "bulk" `read/write_registers` methods". Is that the correct interpretation of your question?

If so. then that is certainly an interesting point, that might merit a separate discussion. I can think of a few reasons why the current approach is better, though on the flipside, it would be nice to simplify this bit of the API...

To avoid burying this discussion in this PR, could you re-state the question as a Discussion thread? That will make it easier to refer back to the discussion in the future.

* * *

> If I do this, I need to replace all the `write_byte` in `gdb_serialize` with copy the data to different part of buffer. Is there any easy way to do this? I may need something like this [`put_u8`](https://docs.rs/bytes/1.1.0/bytes/struct.BytesMut.html#method.put_u8).

I would probably be alright with taking a dependency on something like [`byteorder`](https://docs.rs/byteorder/1.4.3/byteorder/), which includes the functionality you're looking for.

That said, I just brushed up on why exactly I opted to use a callback in `gdb_serialize`, and it was because I wanted to avoid leaking the detail of "unavailable bytes" (i.e: returning "xx") to the `Arch` implementation. Ideally, I'd like to preserve this property in whatever refactored implementation you come up with.

Also, to avoid losing this context and discussion, can you open a new Discussion thread on the topic of refactoring `gdb_serialize`? This isn't something that would land as part of this PR, so there's no reason to have the discussion here.

src/gdbstub_impl/ext/single_register_access.rs
```diff
@@ -13,17 +13,27 @@ impl<T: Target, C: Connection> GdbStubImpl<T, C> {
...
                 let (reg_id, reg_size) = match reg {
                     // empty packet indicates unrecognized query
                     None => return Ok(HandlerStatus::Handled),
                     Some(v) => v,
                 };
-                let dst = &mut dst[0..reg_size];
-                ops.read_register(id, reg_id, dst).handle_error()?;
+                let mut buf = p.buf;
+                if let Some(size) = reg_size {
+                    buf = &mut buf[0..size.get()];
```
daniel5151, line 24: This needs to be checked buffer access, or else this will introduce a panic.

```rus
buf = buf.get_mut(..size.get()).ok_or(Error::PacketBufferOverflow)?;
```

ditto for the buffer access on line 34.

src/target/ext/base/single_register_access.rs
```diff
@@ -22,16 +22,18 @@ pub trait SingleRegisterAccess<Id>: Target {
     /// On single threaded targets, `tid` is set to `()` and can be ignored.
     ///
     /// Implementations should write the value of the register using target's
-    /// native byte order in the buffer `dst`.
+    /// native byte order in the buffer `buf`.
     ///
     /// If the requested register could not be accessed, an appropriate
     /// non-fatal error should be returned.
+    ///
+    /// Return the number of bytes written into `buf`.
```
daniel5151, line 30: nit: move this line right after "Implementations should write the value of the register using target's native byte order in the buffer `buf`"

daniel5151 (changes requested):

bet4it:
> though now that I re-read your question, I think you're actually suggesting something else: "why not implement the `g`/`G` packets via the `read/write_register` methods in the user-level API, removing the current "bulk" `read/write_registers` methods". Is that the correct interpretation of your question?

Yes, that's what I mean!
I will re-state this to Discussion in next days.

> This isn't something that would land as part of this PR, so there's no reason to have the discussion here.

Oh, I just thought that pass `PacketBuf` to `gdb_serialize` is also something related to "Pass PacketBuf when read registers", so I prepared to make the changes in this PR😅

daniel5151:
Lets keep this PR focused on changing the `read_register` API, and the underlying `RegId` API.

I've also gone ahead and changed the PR's title to reflect this focus.

src/arch.rs
```diff
@@ -15,7 +15,7 @@
 //! > Having community-created `Arch` implementations distributed in a separate
 //! crate helps minimize any unnecessary "version churn" in `gdbstub` core.
 
-use core::fmt::Debug;
+use core::{fmt::Debug, num::NonZeroUsize};
```
daniel5151, line 18: nit: imports should only be merged at the "leaf" level:

```suggestion
use core::fmt::Debug;
use core::num::NonZeroUsize;
```

daniel5151 (changes requested):
Tiny nit, but aside from that, I think this is good to merge.

daniel5151 (approved):
Give me a 👍 if this is good to merge :)

PR #87, opened 2021-09-27
Change rip size to 8 in `X86_64CoreRegId`.

### LINKED ISSUE #80: RIP size on x86-64

gz:
### API Stability

Fixes #80.

### Checklist

I ran this in the past but this change was with the github editor :)

### DIFF
--- a/gdbstub_arch/src/x86/reg/id.rs
+++ b/gdbstub_arch/src/x86/reg/id.rs
@@ -172,7 +172,7 @@ impl RegId for X86_64CoreRegId {
 
         let r = match id {
             0..=15 => (Gpr(id as u8), 8),
-            16 => (Rip, 4),
+            16 => (Rip, 8),
             17 => (Eflags, 8),
             18..=23 => (Segment(X86SegmentRegId::from_u8(id as u8 - 18)?), 4),
             24..=31 => (St(id as u8 - 24), 10),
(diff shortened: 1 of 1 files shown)


daniel5151:
Seems like `Eflags` is also incorrect? Shouldn't it be `4`?

daniel5151:
Ah, and it seems you'll need to update the test as well.

gz:
Ha, totally broke the tests with a change that I figured would never do that :) I'm gonna fix it..

Regarding eflags, on x86-64 eflags was technically renamed to rflags and extended to be 64bit (even though the upper 32 bits are not really used):
https://docs.rs/x86/0.42.0/src/x86/bits64/rflags.rs.html#14

I will check the xml of gdb as well to see if it's really 8 bits there.

gz:
Hm seems to be still 32bit in gdb: https://github.com/bminor/binutils-gdb/blob/master/gdb/features/i386/64bit-core.c#L49

So I can change this too.

daniel5151:
Haha, all good.

Regardless what the x86 spec says, we'll need to match GDB's target XML. If they're only using 32 bits, then we should match that behavior.

Of course, if you want to use the full 64-bit value, you can always define a custom `Arch` + `target.xml` that uses the full 64 bits.

gz:
Fixed tests, let me know if this is good.

daniel5151:
Nice, thanks for the fix 🎉

PR #88, opened 2021-09-30
Rewrite `GdbStub::run` to use the state machine implementation (and more!)

daniel5151:
This is a set of API changes I've been sitting on for many months now, and I think it's high-time that I pushed these changes up.

The docs are still a mess, but I've opted to commit things up as is to get hands-on feedback from folks to make sure things work, and then I'll tackle the documentation later.

At a high level, this PR introduces the following API tweaks:

#### Removed the `read` and `peek` methods from `Connection`, and moves them into a separate `ConnectionExt` trait.

The state machine API was explicitly designed to _circumvent_ the need for a blocking `read` API. Nonetheless, it is useful to have a reference implementation of a blocking single-byte read for various types (e.g: TcpStream) built-in to `gdbstub`, so I've split these implementations off into a separate `ConnectionExt` trait - which as the name implies, requires the type to also implement `Connection`.

#### `Target::resume` is now _non blocking_, and does not support immediately returning a `StopReason`

The previous iteration of the API kept the ability to return a `StopReason` from `Target::resume` in order to maintain compatibility with existing blocking `Target::resume` implementations (such as the one in the `armv4t` example).

With this new API, targets are now _forced_ to resume in a non-blocking manner, with the responsibility of "driving" the resumed target lifted up into the state machine event loop.

This change also means that `Target::resume` no longer takes a `GdbInterrupt` callback, which has simplified a _ton_ of code within `gdbstub`!

#### The bespoke blocking `GdbStub::run` implementation has been swapped out with a new `GdbStub::run_blocking` method - backed by the state machine API.

Instead of maintaining two parallel "front-ends" to `GdbStub`, I've consolidated the two to both use the `GdbStubStateMachine` API under the hood. This new API is _looks_ more complex to the old `GdbStub::run` API, as it requires the user to provide an implementation of `gdbstub_run_blocking::BlockingEventLoop` for their particular `Target` + `Connection`, but in reality, all this new API has done is "lift" some of the code that used to live in `Target::resume` out up into `gdbstub_run_blocking::BlockingEventLoop`.


daniel5151:
Pinging @gz and @bet4it

Given that both of you are currently actively working with `gdbstub`, could you please give these changes a try, and validate that they work + make sense?

In the past, I would have simply merged these changes into `dev/0.6` without consulting anyone, but given that these are pretty large changes, I was hoping to get some feedback before I pulled the trigger on them 😅

src/connection/mod.rs
```diff
@@ -61,17 +50,22 @@ pub trait Connection {
...
 /// for [`TcpStream`](std::net::TcpStream) and
 /// [`UnixStream`](std::os::unix::net::UnixStream) (on unix systems).
+///
+/// [`gdbstub_run::Callbacks::read_byte`]:
+/// crate::gdbstub_run::Callbacks::read_byte
 pub trait ConnectionExt: Connection {
     /// Read a single byte.
     fn read(&mut self) -> Result<u8, Self::Error>;
+
+    /// Peek a single byte. This MUST be a **non-blocking** operation, returning
```
gz, line 68: FWIW it might be good to clarify in the docs here how peek should behave together with read. I assume if I do `peak -> read` or `peak -> peak -> read` it will return the "same byte" 2x in the first example and 3x in the second one? 
I 
was mainly wondering about this when `read()` was part of ConnectionExt but `peak()` was part of Connection what their relationship would be so maybe this is clear now.

gz:
I tried it in my branch and FWIW [my silly test](https://github.com/vmware-labs/node-replicated-kernel/blob/gdbstub2/kernel/tests/integration-test.rs#L1053) still passes and the changes make sense to me.

src/connection/mod.rs
daniel5151: Yep, your intuition was right, though I agree that it never hurts to add more clarification to the docs.

bet4it (approved):
I have migrated my project to this implementation and everything works well.
`resume` is much clearer than before.

PR #92, opened 2021-10-04
Make single-stepping optional (removing ResumeAction)

### LINKED ISSUE #59: Make single-stepping optional
A close reading of the GDB RSP docs suggests that single-stepping is actually an _optional_ feature of the protocol. Emphasis mine:
https://sourceware.org/gdb/current/onlinedocs/gdb/Overview.html#Overview

> At a minimum, a stub is required to support the ‘?’ command to tell GDB the reason for halting, ‘g’ and ‘G’ commands for register access, and the ‘m’ and ‘M’ commands for memory access. Stubs that only control single-threaded targets can implement run control with the ‘c’ (continue) command, **and if the target architecture supports hardware-assisted single-stepping, the ‘s’ (step) command**. Stubs that support multi-threading targets should support the ‘vCont’ command. All other commands are optional.

At first blush, it would seem that making single-stepping an optional feature is actually fairly straightforward: simply tweak the `resume` API to model single-stepping in a manner similar to other optional resume modes (such as optimized range stepping). It would require an API breaking change (removing the `ResumeReason::Step{WithSignal}` variants), but aside from that, it shouldn't be too tricky to implement.

* * *

Unfortunately, making single-stepping optional comes with a hidden footgun: if a GDB user tries to single-step on a target that don't support native single-stepping, the GDB client will instead attempt to use temporary breakpoints to "emulate" single stepping. You can see this behavior by skimming through the source of [`gdb/infrun.c:resume_1`](https://github.com/bminor/binutils-gdb/blob/aeeb758df5a61a323de0095c28b3de1ee7b00db3/gdb/infrun.c#L2188).

The `gdbstub` documentation should make it _very clear_ that supporting optimized single stepping is _highly recommended_, as it is significantly more efficient than relying on continue + temporary breakpoints. Moreover, if the target doesn't support single stepping _or_ breakpoints, then calling `step` won't work at all.

daniel5151:
Closes #59

See the associated issue for context and rationale.

This is definitely a big API change, but one that is important to make. Single stepping should _not_ be required by the base protocol, as while it's trivial to implement on some targets (e.g: emulators), it is _not_ trivial to implement in others.

The GDB client is able to spoof single-stepping by setting temporary breakpoint instructions in the guest. While this is undoubtedly slower than an optimized single-stepping implementation, it has the key benefit of working "out-of-the-box", without any explicit effort required by the `gdbstub` user to get it working.

### DIFF
--- a/example_no_std/src/gdb.rs
+++ b/example_no_std/src/gdb.rs
@@ -1,6 +1,6 @@
 use gdbstub::common::Tid;
 use gdbstub::target;
-use gdbstub::target::ext::base::multithread::{MultiThreadOps, ResumeAction};
+use gdbstub::target::ext::base::multithread::MultiThreadOps;
 use gdbstub::target::{Target, TargetResult};
 
 use crate::print_str::print_str;
@@ -52,8 +52,12 @@ impl MultiThreadOps for DummyTarget {
     }
 
     #[inline(never)]
-    fn set_resume_action(&mut self, _tid: Tid, _action: ResumeAction) -> Result<(), Self::Error> {
-        print_str("> set_resume_action");
+    fn set_resume_action_continue(
+        &mut self,
+        _tid: Tid,
+        _signal: Option<u8>,
+    ) -> Result<(), Self::Error> {
+        print_str("> set_resume_action_continue");
         Ok(())
     }
 
--- a/examples/armv4t/gdb/mod.rs
+++ b/examples/armv4t/gdb/mod.rs
@@ -2,9 +2,7 @@ use core::convert::{TryFrom, TryInto};
 
 use armv4t_emu::{reg, Memory};
 use gdbstub::target;
-use gdbstub::target::ext::base::singlethread::{
-    ResumeAction, SingleThreadOps, SingleThreadReverseContOps, SingleThreadReverseStepOps,
-};
+use gdbstub::target::ext::base::singlethread::SingleThreadOps;
 use gdbstub::target::{Target, TargetError, TargetResult};
 use gdbstub_arch::arm::reg::id::ArmCoreRegId;
 
@@ -136,7 +134,7 @@ impl Target for Emu {
 }
 
 impl SingleThreadOps for Emu {
-    fn resume(&mut self, action: ResumeAction) -> Result<(), Self::Error> {
+    fn resume(&mut self, signal: Option<u8>) -> Result<(), Self::Error> {
         // Upon returning from the `resume` method, the target being debugged should be
         // configured to run according to whatever resume actions the GDB client has
         // specified (as specified by `set_resume_action`, `resume_range_step`,
@@ -150,11 +148,11 @@ impl SingleThreadOps for Emu {
         // external "orchestration" to set it's execution mode (e.g: modifying the
         // target's process state via platform specific debugging syscalls).
 
-        self.exec_mode = match action {
-            ResumeAction::Continue => ExecMode::Continue,
-            ResumeAction::Step => ExecMode::Step,
-            _ => return Err("cannot resume with signal"),
-        };
+        if signal.is_some() {
+            return Err("no support for continuing with signal");
+        }
+
+        self.exec_mode = ExecMode::Continue;
 
         Ok(())
     }
@@ -216,20 +214,43 @@ impl SingleThreadOps for Emu {
     }
 
     #[inline(always)]
-    fn support_reverse_cont(&mut self) -> Option<SingleThreadReverseContOps<Self>> {
+    fn support_reverse_cont(
+        &mut self,
+    ) -> Option<target::ext::base::singlethread::SingleThreadReverseContOps<Self>> {
         Some(self)
     }
 
     #[inline(always)]
-    fn support_reverse_step(&mut self) -> Option<SingleThreadReverseStepOps<Self>> {
+    fn support_reverse_step(
+        &mut self,
+    ) -> Option<target::ext::base::singlethread::SingleThreadReverseStepOps<Self>> {
         Some(self)
     }
 
     #[inline(always)]
-    fn support_resume_range_step(
+    fn support_single_step(
+        &mut self,
+    ) -> Option<target::ext::base::singlethread::SingleThreadSingleStepOps<Self>> {
+        Some(self)
+    }
+
+    #[inline(always)]
+    fn support_range_step(
         &mut self,
     ) -> Option<target::ext::base::singlethread::SingleThreadRangeSteppingOps<Self>> {
         Some(self)
+    }
+}
+
+impl target::ext::base::singlethread::SingleThreadSingleStep for Emu {
+    fn step(&mut self, signal: Option<u8>) -> Result<(), Self::Error> {
+        if signal.is_some() {
+            return Err("no support for stepping with signal");
+        }
+
+        self.exec_mode = ExecMode::Step;
+
+        Ok(())
     }
 }
 
--- a/examples/armv4t_multicore/gdb.rs
+++ b/examples/armv4t_multicore/gdb.rs
@@ -2,7 +2,7 @@ use armv4t_emu::{reg, Memory};
 
 use gdbstub::common::Tid;
 use gdbstub::target;
-use gdbstub::target::ext::base::multithread::{MultiThreadOps, ResumeAction};
+use gdbstub::target::ext::base::multithread::MultiThreadOps;
 use gdbstub::target::ext::breakpoints::WatchKind;
 use gdbstub::target::{Target, TargetError, TargetResult};
 
@@ -62,14 +62,24 @@ impl MultiThreadOps for Emu {
         Ok(())
     }
 
-    fn set_resume_action(&mut self, tid: Tid, action: ResumeAction) -> Result<(), Self::Error> {
-        match action {
-            ResumeAction::Step => self.exec_mode.insert(tid_to_cpuid(tid)?, ExecMode::Step),
-            ResumeAction::Continue => self
-                .exec_mode
-                .insert(tid_to_cpuid(tid)?, ExecMode::Continue),
-            _ => return Err("no support for resuming with signal"),
-        };
+    #[inline(always)]
+    fn support_single_step(
+        &mut self,
+    ) -> Option<target::ext::base::multithread::MultiThreadSingleStepOps<Self>> {
+        Some(self)
+    }
+
+    fn set_resume_action_continue(
+        &mut self,
+        tid: Tid,
+        signal: Option<u8>,
+    ) -> Result<(), Self::Error> {
+        if signal.is_some() {
+            return Err("no support for continuing with signal");
+        }
+
+        self.exec_mode
+            .insert(tid_to_cpuid(tid)?, ExecMode::Continue);
 
         Ok(())
     }
@@ -150,6 +160,18 @@ impl MultiThreadOps for Emu {
     ) -> Result<(), Self::Error> {
         register_thread(cpuid_to_tid(CpuId::Cpu));
         register_thread(cpuid_to_tid(CpuId::Cop));
+        Ok(())
+    }
+}
+
+impl target::ext::base::multithread::MultiThreadSingleStep for Emu {
+    fn set_resume_action_step(&mut self, tid: Tid, signal: Option<u8>) -> Result<(), Self::Error> {
+        if signal.is_some() {
+            return Err("no support for stepping with signal");
+        }
+
+        self.exec_mode.insert(tid_to_cpuid(tid)?, ExecMode::Step);
+
         Ok(())
     }
 }
--- a/src/gdbstub_impl/ext/base.rs
+++ b/src/gdbstub_impl/ext/base.rs
@@ -4,7 +4,7 @@ use crate::protocol::commands::ext::Base;
 use crate::arch::{Arch, Registers};
 use crate::protocol::{IdKind, SpecificIdKind, SpecificThreadId};
 use crate::target::ext::base::multithread::ThreadStopReason;
-use crate::target::ext::base::{BaseOps, ReplayLogPosition, ResumeAction};
+use crate::target::ext::base::{BaseOps, ReplayLogPosition};
 use crate::{FAKE_PID, SINGLE_THREAD_TID};
 
 impl<T: Target, C: Connection> GdbStubImpl<T, C> {
@@ -298,13 +298,25 @@ impl<T: Target, C: Connection> GdbStubImpl<T, C> {
                 use crate::protocol::commands::_vCont::vCont;
                 match cmd {
                     vCont::Query => {
-                        res.write_str("vCont;c;C;s;S")?;
+                        // Continue is part of the base protocol
+                        res.write_str("vCont;c;C")?;
+
+                        // Single stepping is optional
                         if match target.base_ops() {
-                            BaseOps::SingleThread(ops) => ops.support_resume_range_step().is_some(),
+                            BaseOps::SingleThread(ops) => ops.support_single_step().is_some(),
+                            BaseOps::MultiThread(ops) => ops.support_single_step().is_some(),
+                        } {
+                            res.write_str(";s;S")?;
+                        }
+
+                        // Range stepping is optional
+                        if match target.base_ops() {
+                            BaseOps::SingleThread(ops) => ops.support_range_step().is_some(),
                             BaseOps::MultiThread(ops) => ops.support_range_step().is_some(),
                         } {
                             res.write_str(";r")?;
                         }
+
                         HandlerStatus::Handled
                     }
                     vCont::Actions(actions) => self.do_vcont(target, actions)?,
@@ -312,23 +324,11 @@ impl<T: Target, C: Connection> GdbStubImpl<T, C> {
             }
             // TODO?: support custom resume addr in 'c' and 's'
             //
-            // unfortunately, this wouldn't be a particularly easy thing to implement, since the
-            // vCont packet doesn't natively support custom resume addresses. This leaves a few
-            // options for the implementation:
+            // vCont doesn't have a notion of "resume addr", and since the implementation of these
+            // packets reuse vCont infrastructure, supporting this obscure feature will be a bit
+            // annoying...
             //
-            // 1. Adding new ResumeActions (i.e: ContinueWithAddr(U) and StepWithAddr(U))
-            // 2. Automatically calling `read_registers`, updating the `pc`, and calling
-            //    `write_registers` prior to resuming.
-            //    - will require adding some sort of `get_pc_mut` method to the `Registers` trait.
-            //
-            // Option 1 is easier to implement, but puts more burden on the implementor. Option 2
-            // will require more effort to implement (and will be less performant), but it will hide
-            // this protocol wart from the end user.
-            //
-            // Oh, one more thought - there's a subtle pitfall to watch out for if implementing
-            // Option 1: if the target is using conditional breakpoints, `do_vcont` has to be
-            // modified to only pass the resume with address variants on the _first_ iteration
-            // through the loop.
+            // TODO: add `support_legacy_s_c_packets` flag (similar to `use_X_packet`)
             Base::c(_) => {
                 use crate::protocol::commands::_vCont::Actions;
 
@@ -474,28 +474,41 @@ impl<T: Target, C: Connection> GdbStubImpl<T, C> {
             return Err(Error::PacketUnexpected);
         }
 
-        let action = match first_action.kind {
-            VContKind::Step => ResumeAction::Step,
-            VContKind::Continue => ResumeAction::Continue,
-            VContKind::StepWithSig(sig) => ResumeAction::StepWithSignal(sig),
-            VContKind::ContinueWithSig(sig) => ResumeAction::ContinueWithSignal(sig),
-            VContKind::RangeStep(start, end) => {
-                if let Some(ops) = ops.support_resume_range_step() {
-                    let start = start.decode().map_err(|_| Error::TargetMismatch)?;
-                    let end = end.decode().map_err(|_| Error::TargetMismatch)?;
+        match first_action.kind {
+            VContKind::Continue | VContKind::ContinueWithSig(_) => {
+                let signal = match first_action.kind {
+                    VContKind::ContinueWithSig(sig) => Some(sig),
+                    _ => None,
+                };
 
-                    ops.resume_range_step(start, end)
-                        .map_err(Error::TargetError)?;
-                    return Ok(());
-                } else {
-                    return Err(Error::PacketUnexpected);
-                }
+                ops.resume(signal).map_err(Error::TargetError)?;
+                Ok(())
+            }
+            VContKind::Step | VContKind::StepWithSig(_) if ops.support_single_step().is_some() => {
+                let ops = ops.support_single_step().unwrap();
+
+                let signal = match first_action.kind {
+                    VContKind::StepWithSig(sig) => Some(sig),
+                    _ => None,
+                };
+
+                ops.step(signal).map_err(Error::TargetError)?;
+                Ok(())
+            }
+            VContKind::RangeStep(start, end) if ops.support_range_step().is_some() => {
+                let ops = ops.support_range_step().unwrap();
+
+                let start = start.decode().map_err(|_| Error::TargetMismatch)?;
+                let end = end.decode().map_err(|_| Error::TargetMismatch)?;
+
+                ops.resume_range_step(start, end)
+                    .map_err(Error::TargetError)?;
+                Ok(())
             }
             // TODO: update this case when non-stop mode is implemented
-            VContKind::Stop => return Err(Error::PacketUnexpected),
-        };
-
-        ops.resume(action).map_err(Error::TargetError)
+            VContKind::Stop => Err(Error::PacketUnexpected),
+            _ => Err(Error::PacketUnexpected),
+        }
     }
 
     fn do_vcont_multi_thread(
@@ -514,49 +527,69 @@ impl<T: Target, C: Connection> GdbStubImpl<T, C> {
                 crate::protocol::PacketParseError::MalformedCommand,
             ))?;
 
-            let resume_action = match action.kind {
-                VContKind::Step => ResumeAction::Step,
-                VContKind::Continue => ResumeAction::Continue,
-                // there seems to be a GDB bug where it doesn't use `vCont` unless
-                // `vCont?` returns support for resuming with a signal.
-                VContKind::StepWithSig(sig) => ResumeAction::StepWithSignal(sig),
-                VContKind::ContinueWithSig(sig) => ResumeAction::ContinueWithSignal(sig),
-                VContKind::RangeStep(start, end) => {
-                    if let Some(ops) = ops.support_range_step() {
-                        match action.thread.map(|thread| thread.tid) {
-                            // An action with no thread-id matches all threads
-                            None | Some(SpecificIdKind::All) => {
-                                return Err(Error::PacketUnexpected)
-                            }
-                            Some(SpecificIdKind::WithId(tid)) => {
-                                let start = start.decode().map_err(|_| Error::TargetMismatch)?;
-                                let end = end.decode().map_err(|_| Error::TargetMismatch)?;
+            match action.kind {
+                VContKind::Continue | VContKind::ContinueWithSig(_) => {
+                    let signal = match action.kind {
+                        VContKind::ContinueWithSig(sig) => Some(sig),
+                        _ => None,
+                    };
 
-                                ops.set_resume_action_range_step(tid, start, end)
-                                    .map_err(Error::TargetError)?;
-                                continue;
-                            }
-                        };
-                    } else {
-                        return Err(Error::PacketUnexpected);
+                    match action.thread.map(|thread| thread.tid) {
+                        // An action with no thread-id matches all threads
+                        None | Some(SpecificIdKind::All) => {
+                            // Target API contract specifies that the default
+                            // resume action for all threads is continue.
+                        }
+                        Some(SpecificIdKind::WithId(tid)) => ops
+                            .set_resume_action_continue(tid, signal)
+                            .map_err(Error::TargetError)?,
                     }
+                }
+                VContKind::Step | VContKind::StepWithSig(_)
+                    if ops.support_single_step().is_some() =>
+                {
+                    let ops = ops.support_single_step().unwrap();
+
+                    let signal = match action.kind {
+                        VContKind::StepWithSig(sig) => Some(sig),
+                        _ => None,
+                    };
+
+                    match action.thread.map(|thread| thread.tid) {
+                        // An action with no thread-id matches all threads
+                        None | Some(SpecificIdKind::All) => {
+                            error!("GDB client sent 'step' as default resume action");
+                            return Err(Error::PacketUnexpected);
+                        }
+                        Some(SpecificIdKind::WithId(tid)) => {
+                            ops.set_resume_action_step(tid, signal)
+                                .map_err(Error::TargetError)?;
+                        }
+                    };
+                }
+
+                VContKind::RangeStep(start, end) if ops.support_range_step().is_some() => {
+                    let ops = ops.support_range_step().unwrap();
+
+                    match action.thread.map(|thread| thread.tid) {
+                        // An action with no thread-id matches all threads
+                        None | Some(SpecificIdKind::All) => {
+                            error!("GDB client sent 'range step' as default resume action");
+                            return Err(Error::PacketUnexpected);
+                        }
+                        Some(SpecificIdKind::WithId(tid)) => {
+                            let start = start.decode().map_err(|_| Error::TargetMismatch)?;
+                            let end = end.decode().map_err(|_| Error::TargetMismatch)?;
+
+                            ops.set_resume_action_range_step(tid, start, end)
+                                .map_err(Error::TargetError)?;
+                        }
+                    };
                 }
                 // TODO: update this case when non-stop mode is implemented
                 VContKind::Stop => return Err(Error::PacketUnexpected),
-            };
-
-            match action.thread.map(|thread| thread.tid) {
-                // An action with no thread-id matches all threads
-                None | Some(SpecificIdKind::All) => {
-                    if !matches!(resume_action, ResumeAction::Continue) {
-                        error!("GDB client sent default resume action that wasn't 'Continue'");
-                        return Err(Error::PacketUnexpected);
-                    }
-                }
-                Some(SpecificIdKind::WithId(tid)) => ops
-                    .set_resume_action(tid, resume_action)
-                    .map_err(Error::TargetError)?,
-            };
+                _ => return Err(Error::PacketUnexpected),
+            }
         }
 
         ops.resume().map_err(Error::TargetError)
--- a/src/target/ext/base/mod.rs
+++ b/src/target/ext/base/mod.rs
@@ -20,29 +20,6 @@ pub enum BaseOps<'a, A, E> {
     MultiThread(&'a mut dyn multithread::MultiThreadOps<Arch = A, Error = E>),
 }
 
-/// Describes how the target should be resumed.
-///
-/// Due to a quirk / bug in the mainline GDB client, targets are required to
-/// handle the `WithSignal` variants of `Step` and `Continue` regardless of
-/// whether or not they have a concept of "signals".
-///
-/// If your target does not support signals (e.g: the target is a bare-metal
-/// microcontroller / emulator), the recommended behavior is to either return a
-/// target-specific fatal error, or to handle `{Step,Continue}WithSignal` the
-/// same way as their non-`WithSignal` variants.
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
-pub enum ResumeAction {
-    /// Continue execution, stopping once a
-    /// [`StopReason`](singlethread::StopReason) occurs.
-    Continue,
-    /// Step execution.
-    Step,
-    /// Continue with signal.
-    ContinueWithSignal(u8),
-    /// Step with signal.
-    StepWithSignal(u8),
-}
-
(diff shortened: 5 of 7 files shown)


daniel5151:
@gz and @bet4it, could I ask you two to do me a favor and make sure these changes work as expected?

I've confirmed that things are working in the `armv4t` and `armv4t_multicore` examples, but it never hurts to get an extra set of eyes on these kinds of large changes!

daniel5151:
The bug is now filed at https://sourceware.org/bugzilla/show_bug.cgi?id=28440

Lets hope it hasn't actually been sent to `/dev/null` like some of the other bugs/questions I've reported...

bet4it:
I finally have time to figure out the reason why GDB client doesn't respect the optional nature of single-stepping on certain architectures:

GDB parses the result of `vCont?` in `remote_target::remote_vcont_probe`:
https://github.com/bminor/binutils-gdb/blob/c599303f92b1edd6ead3947737a8ae9e1c85e08c/gdb/remote.c#L6117-L6167
The state of `s` and `S` is stored in `rs->supports_vCont.s` and `rs->supports_vCont.S` and is only used in `remote_target::can_do_single_step`:
https://github.com/bminor/binutils-gdb/blob/c599303f92b1edd6ead3947737a8ae9e1c85e08c/gdb/remote.c#L14316-L14335
And if you search `can_do_single_step` in GDB source code you will find that it only be used in ARM:
https://github.com/bminor/binutils-gdb/blob/c599303f92b1edd6ead3947737a8ae9e1c85e08c/gdb/arm-linux-tdep.c#L927

That means that, although we *guess* from the spec ( I don't find the clear explanation of `vCont?` ) that GDB client should respect the result of `vCont?`, but the implementation of GDB itself doesn't really care about it ( except in ARM ).
If an architecture wants to support software single step, it need to register the function with `set_gdbarch_software_single_step`. You may find that some architectures such as X86 doesn't set it, so it's impossible to use software single step on such architectures.

The result of above: only the behavior of ARM will be affected by `vCont?`. Architectures such as X86 and AArch64 will always use hardware single step, and architectures such as MIPS will always use software single step, despite the result of `vCont?`. I have tested and confirmed it.

Then the code currently should be changed:
1. We use `supports_optional_single_step` to make sure single step is implemented on X86 so the `step` command can work correctly ( although `step` could not be used during one debug session ). If you think it's necessary, then we may also need to make sure the user implement software breakpoint on architectures such as MIPS, otherwise we can't do `step`.
2. Or I think we could just remove the guard rail of optional single step and let the user choose which is necessary to implemented.

daniel5151:
Thanks for the deep-dive @bet4it. This is super valuable stuff.

Indeed, the idea behind the `supports_optional_single_step` architecture is to be able to encode whether the GDB client exhibits this issue on a per-arch level. At the time, we'd only confirmed that x86 requires hardware single stepping, and that ARM does not, falling back to the known-safe baseline behavior of "you must support single-step" on all other architectures (that we did not hands-on audit).

> That means that, although we _guess_ from the spec ( I don't find the clear explanation of `vCont?` ) that GDB client should respect the result of `vCont?`, but the implementation of GDB itself doesn't really care about it ( except in ARM ).

Yep, this behavior is definitely not "in the spec", and is strictly an issue with the GDB client implementation. Unfortunately, it's enough of a usability concern that it makes sense for `gdbstub` to include a guard rail to make sure end users aren't bitten by any surprising errors (as gz was when he was working on his implementation).

> If you think it's necessary, then we may also need to make sure the user implement software breakpoint on architectures such as MIPS, otherwise we can't do `step`.

As is precedent in the project, I would not be comfortable encoding these sorts of invariants without getting sign-off from someone with a working target implementation for a particular architecture. i.e: I'll only merge a arch-level `supports_optional_single_step` update if the author has validated the behavior on a "real" target implementation.

To reiterate though: the baseline behavior of the `supports_optional_single_step` infrastructure is to _enforce_ targets implementing single-step support, unless they have been explicitly tested to support implicit support (i.e: ARM), so your example with MIPS is currently doing the right thing.

* * *

What I should do is spin up a tracking issue that tracks which current Arch imlpls in `gdbstub_arch` still need to be audited. Maybe I'll do that sometime today...

daniel5151:
Also, if you don't mind, could you post that exploration over on the upstream bug report at https://sourceware.org/bugzilla/show_bug.cgi?id=28440?

bet4it:
I think it's not a bug of GDB.
Firstly, the spec never guarantees when `vCont?` doesn't get `;s;S`, GDB will use breakpoints to emulate single stepping. You get this illusion by the behaviour on ARM and guessing of the thing `vCont?` should do.
Then it's no problem that the implementation only covers part of the spec, right? 99.99% users of GDB client connect to gdbserver that running on a real hardware, and whether the CPU architecture supports hardware single step is deterministic. If the architecture don't support hardware single step, does GDB need to add code about hardware single step? ( although it's still possible when we emulate the architecture with a software way ) And if the architecture support hardware single step, it's not necessary for GDB to emulate single stepping by breakpoints ( although GDB does this on ARM ). If GDB fully considers the result of `;s;S`, it could even not find a place to test the code ( where can you find a MIPS CPU that supports hardware single step? ), and nearly no people will use this function.

And you did't understand what I meant before. You think step can be optional because if it's not implemented, GDB should emulate it by temporary breakpoints. But I think step can be optional because if the user don't use commands like `stepi`, step won't be used, so we can just leave it to the user, let the user choose if they need to implement it.
You use `supports_optional_single_step` to make sure step can still work on some architectures even if the user don't implement it. But why step is so important that it must work? What if breakpoint is not implemented neither? And as I said before, only ARM will use breakpoint as a backup way of step.

> to make sure end users aren't bitten by any surprising errors

According to common sense, the user shouldn't expect step can work if they not implement it ( Yes ARM can give us a suprise but it's not the common behaviour ). And if a user uses gdbstub on MIPS, he will find that step still can't work even if he implements the step function, because GDB will only use breakpoints to implement single step on such architectures, and doesn't consider if you support step by yourself. You use `supports_optional_single_step` to make sure step can work, then do we need another guard rail on MIPS to enforce the user implement breakpoint so step can work? I think it's better to just remove all the guard rails.

> I would not be comfortable encoding these sorts of invariants without getting sign-off from someone with a working target implementation for a particular architecture.

I'm working on a project that uses `gdbstub` and can work on all major architectures ( X86, X86_64, ARM, AArch64, MIPS currently and can support PowerPC and RISCV with small changes later ). It will be public later ( I really hope gdbstub can release `0.6` before I release my project ), but if you want to test different architectures by yourself, I can give you privilege to my project.

daniel5151:
> I think it's not a bug of GDB.

This is definitely a bug in GDB. If you boil it down to the bare minimum, the bug is very simple: if a client doesn't respond with support for `s;S`, _the GDB client should not send `s` or `S` or `vCont;s;S` packets!_ 

The issue at hand is that when the target doesn't report support for single stepping, and the user tries to use `stepi`, the client doesn't report "command not supported by remote target" (as it would for any other packet that the target hasn't specified support for), and on some architectures, it simply sends the packet anyways.

In other words, the GDB client is not respecting the optional nature of single stepping, as specified by the GDB spec, and it needs to be updated to either a) force remote targets on certain arch's (e.g: x86) to support single-stepping, or b) properly report "stepi is not supported" when the target doesn't respond with support for single-stepping.

> You use `supports_optional_single_step` to make sure step can work

That's not true at all. The purpose of this guard rail is to work around the aforementioned GDB bug, and avoid the `gdbstub` unexpectedly dying due to "packet unsupported" errors when the user attempts to `stepi` on a a target that hasn't implemented the single-stepping extension.

The reason I've made this guard rail configurable on a per-arch level is that the GDB client clearly has the ability to work around the lack of single-stepping support on certain platforms (e.g: ARM), and it would be a shame to avoid exposing that functionality.

> I'm working on a project that uses `gdbstub` and can work on all major architectures ( X86, X86_64, ARM, AArch64, MIPS currently and can support PowerPC and RISCV with small changes later ).

Ah, nice! Based on our previous interactions, I had assumed you were working with primarily x86. 

Sure, if you've got an implementation that works across multiple architectures, I'd me more than happy to take your word on how various features work on a certain arch.

e.g: if you have the time, it would be great if you could check whether other platforms exhibit the aforementioned bug, so that we can update their archs' `supports_optional_single_step` accordingly :)

* * *

> And if a user uses gdbstub on MIPS, he will find that step still can't work even if he implements the step function, because GDB will only use breakpoints to implement single step on such architectures, and doesn't consider if you support step by yourself.

I don't have access to a MIPS target, so I can't double check that behavior, nor do I fully understand what behavior you're describing here...

I interpret what you wrote as as "the GDB client doesn't use the single-step handler on MIPS targets, even if its present". 

If that's the right interpretation, then that's not _great_, but it's also not something `gdbstub` really needs to care about. i.e: the bug would manifest as "single stepping isn't as fast as it could be", rather than "`gdbstub` crashes due to the client sending an unexpected packet".

Though it might be worthwhile to document this behavior somewhere anyways...

* * *

> According to common sense, the user shouldn't expect step can work if they not implement it

> I think it's better to just remove all the guard rails.

You really need to get out of the C/C++ programmer mindset! 😄 

Rust the language, and `gdbstub` the project, are all about adding guard rails! 
e.g: array indexes in Rust are bounds checked by default because time has shown _very clearly_ that programmers are human, and don't always write perfect code.
e.g: `gdbstub`'s entire API is structured around IDETs to enforce that certain "groups" of packets are implemented together (consider `z` and `Z`).

The ethos of `gdbstub`, as well as any idiomatic Rust project, is to make it _as hard as possible_ to misuse the APIs being exposed, while also providing an "escape hatch" for users that know what they're doing.

e.g: with checked array indexing, Rust has a guard rail for most casual users, ensuring they don't accidentally corrupt their own memory, while also providing `unsafe` methods for "power users" that give them the ability to override these guard rails when required.

e.g: with `gdbstub`, the various guard rails are in place to ensure users don't run into any unexpected footguns out of the library's control (such as this optional single step bug, or the implicit software breakpoints behavior). Sure, you could mention these behaviors purely in the library's docs, but as time has shown many times, _invariants encoded in doc comments are often skimmed over and forgotten!_

The crux of all these guard rails, API contortions, and extra engineering effort boils down to two very simple words: _user empathy_.

I _hate_ debugging API misuse issues in C/C++ projects, and I appreciate that Rust's language and community is of the opinion that APIs should be as difficulty to misuse as possible. That makes my life a _lot_ easier as a library consumer, and as such, I try and pass on these gains to folks that end up consuming my library.

daniel5151:
Oh, and as for releasing `gdbstub` 0.6: I don't forsee any more breaking functionality being added prior to release. The only blocker now is documentation, which I should have time to do sometime over the next months or two (with the holidays coming up, I should have more time to sit down and hammer stuff out).

That said, if you're squeamish about pulling deps directly from git + tying to a particular commit hash, I don't mind pushing out a `0.6.0-beta-1` release on crates.io in the interim...

bet4it:
Oh no, you totally don't understand me...

> if a client doesn't respond with support for `s;S`, *the GDB client should not send `s` or `S` or `vCont;s;S` packets!*

But the spec doesn't point it out. Don't you think we should always follow the spec and if the spec doesn't say something we should assume that it doesn't exist?

According to the code of GDB, this behaviour does exist, but only on ARM.

And even client responds with support for `s;S`, GDB client may not send `s` or `S` or `vCont;s;S`. GDB client will use breakpoint on MIPS.

> force remote targets on certain arch's (e.g: x86) to support single-stepping

How can GDB do this? For example, how can GDB force remote targets support `vMustReplyEmpty`?

> avoid the `gdbstub` unexpectedly dying due to "packet unsupported" errors when the user attempts to `stepi` on a a target that hasn't implemented the single-stepping extension.

If the GDB server doesn't respond with support for `s;S` on ARM, and breakpoint is not implemented, `gdbstub` will also die unexpectedly when the user attempts to `stepi`. How can you solve it? 

> the GDB client clearly has the ability to work around the lack of single-stepping support on certain platforms (e.g: ARM)

> if you have the time, it would be great if you could check whether other platforms exhibit the aforementioned bug

Didn't I tell you that according to the code and the behaviour, only GDB on ARM has the ability to work around the lack of single-stepping support? This work around is a special case!

> I interpret what you wrote as as "the GDB client doesn't use the single-step handler on MIPS targets, even if its present".

Yes, that's what I want to say.

> but it's also not something `gdbstub` really needs to care about

But the `gdbstub` will die unexpectedly due to "packet unsupported" errors when the user attempts to `stepi` on a target that hasn't implemented the breakpoints extension...

bet4it:
> The only blocker now is documentation, which I should have time to do sometime over the next months or two

Oh no, there should be some small changes we discussed before. For example, [make `resume` optional](https://github.com/daniel5151/gdbstub/issues/59#issuecomment-933096274).

> That said, if you're squeamish about pulling deps directly from git + tying to a particular commit hash, I don't mind pushing out a `0.6.0-beta-1` release on crates.io in the interim..

Oh, it's not necessary. I can just target on the `dev/0.6` branch.

daniel5151:
> But the spec doesn't point it out.

That's the entire point of the `vCont?` packet. It gives the GDB client a chance to perform feature negotiation with the remote target. If the remote target doesn't support a particular resume packet, the GDB client should not send it. Plain and simple.

> > force remote targets on certain arch's (e.g: x86) to support single-stepping
> 
> How can GDB do this?

Apologies, allow me clarify: when I say "force" I mean "report an error". i.e: if you're trying to resume an x86 target that doesn't support single step, the GDB client should simply refuse error when attempting to `stepi`, thereby "forcing" the target implementation to go back and implement support for single-stepping (if that's a feature they want to use)

Again, if you take a step back for a moment, and look at the bigger picture (ignoring the specific cases of x86, ARM, MIPS, etc...), the core bug is simple and undeniable: the GDB client _must_ respect the feature negotiation that happens as part of `vCont?`, either by working around the lack of `stepi` support by falling back to a breakpoint based approach, or simply erroring when the user tries to run `stepi`. 

It should _not_ send _explicitly unsupported_ vCont packets to the remote target after `vCont?` feature negotiation.

> Didn't I tell you that according to the code and the behaviour, only GDB on ARM has the ability to work around the lack of single-stepping support? This work around is a special case!

Careful now, you're doing the thing again! You know, the thing thing where you make some short-term design decisions based on the concrete implementation details of a particular GDB client 👀 

Based on my reading of the spec (and in this case, I'm almost certain that I'm reading it right), _all_ targets _should_ be allowed to omit support for single-stepping (as part of `vCont` feature negotiation), and in those cases, the GDB client should either a) disallow using the `stepi` GDB command, or b) work around the lack of support using breakpoints (if available).

The fact that the current GDB client has a bug, and _doesn't_ follow the spec is _not_ `gdbstub`'s problem. We follow the spec, and anticipate that at some point in the future, GDB clients fix this bug. At that point, `gdbstub` users will have the option to bypass the current guard rail, and have single-step be an optional feature across _all_ architectures.

* * *

> But the `gdbstub` will die unexpectedly due to "packet unsupported" errors when the user attempts to `stepi` on a target that hasn't implemented the breakpoints extension...

Oops, yep, you're totally right, and that's something I'd intended to handle in the guard rail, and simply forgot to do. It's very easy to check for - simply augment the guard rail to check if the target has implemented breakpoint support :)

Thank you for pointing this out - I'll push out a fix for this shortly...

* * *

> Oh no, there should be some small changes we discussed before. For example, [make `resume` optional](https://github.com/daniel5151/gdbstub/issues/59#issuecomment-933096274).

Shoot, I did say that I'd look into that, didn't I...
Okay, fair enough 😅

I'll try and land that change before pushing out `0.6`. As I mentioned in the comment, it should be a mostly mechanical change to split out this functionality from the base method set...

daniel5151:
> > But the `gdbstub` will die unexpectedly due to "packet unsupported" errors when the user attempts to `stepi` on a target that hasn't implemented the breakpoints extension...
> 
> Oops, yep, you're totally right, and that's something I'd intended to handle in the guard rail, and simply forgot to do. It's very easy to check for - simply augment the guard rail to check if the target has implemented breakpoint support :)

After digging into this a bit, I realized that this isn't actually the case. If the target doesn't implement breakpoint packets, the GDB client will _not_ send breakpoint packets, and will instead attempt to overwrite target memory with breakpoint instructions (which would only happen if the user enabled the `use_implicit_sw_breakpoints` guard rail)

daniel5151:
5bc5831b + 1c6560e7 has made all forms of resume optional 🎉

bet4it:
I'm a bit busy recently. Why not you just support more architectures (such as MIPS) in [gdb-optional-step-bug](https://github.com/daniel5151/gdb-optional-step-bug) so you can try it by yourself? I think it's quite simple.

daniel5151:
Hey @bet4it,

To refresh your memory on how the current implementation works:

- `supports_optional_single_step` defaults to `false` on all platforms that don't explicitly support it (which is just ARM right now)
    - This is a reasonable baseline behavior, as the alternative of having it `true` by default would lead to fatal `PacketUnexpected` on platforms that upstream GDB considers single-stepping to be required on (i.e: ignoring the target's `vCont` response entirely), e.g: x86
    - If the two options here are "crash `gdbstub` with an opaque error + trash the user's GDB session" vs. "force the user to implement single-step / manually investigate if their platform requires single-step", I know which one I'm picking
- If the target truly supports optional single step, and _doesn't_ implement SW breakpoint, the `use_implicit_sw_breakpoints` will remind the user that attempting to use single step / continue in GDB without implementing explicit support for sw breakpoints will result in memory being overwritten, which should be enough of a wakeup call to have the user implement sw breakpoints.

AFAICT, these two guard rails should work in tandem to cover the cases you care about (notably, your quibbles about MIPS).

* * *

Please remember that the whole point of these various guard rails is to work around the fact that the mainline GDB client has some edge cases in how it handles targets with "non-standard subsets" of supported features (i.e: spec-compliant, but rarely seen "in the wild"). I'd much rather document + work around these upstream bugs in the context of `gdbstub`, as I don't want folks to run into these issues when working on their `gdbstub` implementation + open bug reports on my project that require upstream patches.

And of course, if you don't like these guard rails - you can just turn them off! Overriding `supports_optional_single_step` to always return `true` lands you right back at option 2 in https://github.com/daniel5151/gdbstub/pull/92#issuecomment-971641019, whereby it'd be up to the GDB client user to make sure they aren't invoking the wrong command.

bet4it:
I want to ask you, what's the meaning of `supports_optional_single_step`?

You use it to represent whether or not the target supports single-stepping as an optional feature.
I guess that in your option if the target supports it (such as `arm`), that means:
1. You can choose if you want to implement single step manually. The behavior of GDB client will change base on it.
2. If you implement single step, then it's obvious that single step can be used. And even if you don't implement single step manually, GDB will emulate it by breakpoint, so it still can be used.

And if the target doesn't support it (such as `x86`), that means:

3. You must implement single step manually to support single step, otherwise `gdbstub` will die unexpectedly due to "packet unsupported" errors.

Is my understanding correctly?

Firstly, as I said before, the spec never guarantees the behavior of GDB client need to change base on the result of `vCont?`, and we can know from the code that only `arm` has the special treatment for this, so this is only a special case, or you can think it as a feature of `arm`. You think it should be very common just because mainly of your work are done on `arm`. I don't think it's appropriate to add a global option on just a feature of one specific architecture.

As for `2` and `3`, I told you that there is another situation: **You must implement breakpoint manually to support single step**, and if you implement single step but no breakpoint the single step feature still can't be used on MIPS. Then what value `supports_optional_single_step` should be set on such architectures?

daniel5151:
So, it does indeed seem that you're misunderstanding the purpose of this guard rail infrastructure. Maybe this is indicative of poor naming or incomplete docs, in which case I'll have to think about what a better name might be...

I _think_ the docs make it clear enough, but maybe you can offer some clarifications / suggestions if the current docs aren't sufficient: https://github.com/daniel5151/gdbstub/blob/3e5d2bc/src/arch.rs#L161-L196

In any case, let me go through your explanation and correct some things:

> You use it to represent whether or not the target supports single-stepping as an optional feature.

So, I think this might be a root of confusion. The intent here is that the value this method returns is dependant on whether or not the _upstream GDB client_ supports optional single step for the architecture, which in turn will guide whether or not the target is allowed to treat single step as optional.

With that in mind...

> 1. You can choose if you want to implement single step manually. The behavior of GDB client will change base on it.

If you implement single step, `gdbstub` will send `s;S` as part of the response to `vCont?`, notifying the GDB client that the target supports single step. If you don't, it will not.

This behaviour has nothing to do with the value of this guard rail.

> 2\. If you implement single step, then it's obvious that single step can be used. And even if you don't implement single step manually, GDB will emulate it by breakpoint, so it still can be used.

So, clarification: GDB will emulate it by breakpoint, but what _kind_ of breakpoint will depend on whether or not sw breakpoints are supported. If a target doesn't implement single step, AND doesn't implement sw breakpoints, then GDB will literally overwrite instruction memory with undefined instructions to insert a "breakpoint". This is very weird behaviour in some contexts (notably, emulators/hypervisors), which is why the `use_implicit_sw_breakpoints` exists. It will loudly complain in this case, and remind the user that they must implement sw breakpoints OR accept the fact that GDB will overwrite memory.

> And if the target doesn't support it (such as `x86`), that means:
> 
> 3. You must implement single step manually to support single step, otherwise `gdbstub` will die unexpectedly due to "packet unsupported" errors.

This is exactly correct. On x86, GDB will NOT use breakpoints (software, or implicit sw) to implement single step, and ALSO ignores the response from `vCont?`, assuming unconditional support for single step. Therefore, the single-step IDET MUST be implemented, or else when the GDB client sends `s` to `gdbstub`, that would be an unexpected packet, and cause a runtime error.

> Firstly, as I said before, the spec never guarantees the behavior of GDB client need to change base on the result of `vCont?`,

Right, well, that's where I fundamentally have to disagree. `vCont?` is _clearly_ a handshake feature negotiation packet, so the correct behaviour of the upstream GDB client is to respect the response from the target. The fact that it doesn't do this for single step on certain (possibly many!) architectures is a _bug_, and _should_ be fixed upstream. 

I suspect the reason no one has reported this in the past is because no one has ever written a truly generic `gdbstub` yet, with each existing implementation either being tied to a particular execution framework (e.g: QEMU), where single stepping is always supported regardless of platform, or the implementation is tied to a single platform/architecture, so supporting single-step is either in or not, without any configurability.


> **You must implement breakpoint manually to support single step**, and if you implement single step but no breakpoint the single step feature still can't be used on MIPS. Then what value `supports_optional_single_step` should be set on such architectures?

So, this is exactly the use case for the other guard rail: `use_implicit_sw_breakpoints`. In this scenario, that other guard rail would fire, and remind you that you need to support some kind of breakpoints, as if you don't, GDB will attempt to emulate them via overriding memory.

In other words, this scenario you're explaining has _nothing to do with optionally supporting single step!_

* * *

Please let me know if that makes sense, and if there's anything you still don't understand. I'm very open to improving the docs to avoid this sort of confusion in the future.

daniel5151:
Alright, it seems you're on the same page as me wrt how I'm thinking about the situation. That's a start.

> But it seems that you don't think about the situation that always implements single step by breakpoints in your ideal GDB client? You don't answer me that what value `supports_optional_single_step` should be set on such situations, or you think that the option is meaningless on such situations so we don't need to set it?

Once again, the value of `supports_optional_single_step` is _independent_ of the particular mechanism by which single stepping is implemented - be it native (e.g: the TRACE flag on x86), hw/sw breakpoints (via the GDB RSP), or implicit (by overriding memory). The purpose of this guard rail to work around the fact that the current mainline upstream GDB client doesn't respect the `s;S` (or lack thereof) in the `vCont?` packet uniformly across all platforms. 

That's it. No more, no less.

I will try to make the docs clearer in this respect, and make it more "front and center" that this guard rail is about working around a mainline GDB client bug.

* * *

> > If GDB fully considers the result of `;s;S`, it could even not find a place to test the code ( where can you find a MIPS CPU that supports hardware single step? ), and nearly no people will use this function.

A few very pertinent points in regards to this one statement:

1. the fact that optional single stepping was only introduced 2 years after `gdbstub`'s first release is precisely _because_ it's an incredibly niche optional feature, and almost _all_ implementation do end up implementing support for single step! I'd really like to end this back and forth, because we're splitting hairs on a feature that most folks will never care about in the first place.

2. In my opinion (and feel free to vehemently disagree), if the GDB client _expects_ a target to support single-step, it must _error out_ if the target doesn't respond with single-step capabilities in response to `vCont?`, as opposed to ignoring this part of the handshake and blindly sending unsupported packets to the target. That's just sloppy software engineering, and is absolutely a bug in my book - hence why I've filed a bug upstream regarding this issue.

3. You're once again conflating single step _implementation_ with the purpose of this guard rail. It doesn't matter if MIPS doesn't support architecture-level _hardware_ single step - the point of this guard rail is to protect against the GDB client unconditionally sending `s` packets on targets that have not implemented support for them. Whether or not it does that on MIPS... I have no idea, I haven't tested it. Someone should check, and if it doesn't then we can loosen the guard rail for the MIPS implementation in `gdbstub_arch`.
    - And to reiterate: this is a guard rail that only really matter when using _off the shelf_ `Arch` implementations from `gdbstub_arch`. If you're already implementing your own `Arch`, then you have free reign over whether this guard rail should be enabled or not.

* * *

> You think guard rail is so important, then we may add some guard rails to force the user to implement breakpoint if he want to support single step on MIPS?

Dude, that is **literally** the current behavior of the `use_implicit_sw_breakpoints` guard rail!! This is what I've been trying to tell you across countless replies?!

If the user implements support for single-step on MIPS (via the `set_resume_action_step` IDET) _without_ implementing any kind of breakpoint support, the `use_implicit_sw_breakpoints` will fire by default to warn them that GDB will attempt to emulate breakpoints by overriding memory. At this point, they can either opt-in to the functionality, or go ahead and implement breakpoints.

This is an explanation I've included _many_ times, and it seems you're just not listening to me:

https://github.com/daniel5151/gdbstub/pull/92#issuecomment-1009607788

> > **You must implement breakpoint manually to support single step**, and if you implement single step but no breakpoint the single step feature still can't be used on MIPS. Then what value `supports_optional_single_step` should be set on such architectures?
> 
> So, this is exactly the use case for the other guard rail: `use_implicit_sw_breakpoints`. In this scenario, that other guard rail would fire, and remind you that you need to support some kind of breakpoints, as if you don't, GDB will attempt to emulate them via overriding memory.

https://github.com/daniel5151/gdbstub/pull/92#issuecomment-1009152843

> * If the target truly supports optional single step, and _doesn't_ implement SW breakpoint, the `use_implicit_sw_breakpoints` will remind the user that attempting to use single step / continue in GDB without implementing explicit support for sw breakpoints will result in memory being overwritten, which should be enough of a wakeup call to have the user implement sw breakpoints.

https://github.com/daniel5151/gdbstub/pull/92#issuecomment-979590493

> After digging into this a bit, I realized that this isn't actually the case. If the target doesn't implement breakpoint packets, the GDB client will _not_ send breakpoint packets, and will instead attempt to overwrite target memory with breakpoint instructions (which would only happen if the user enabled the `use_implicit_sw_breakpoints` guard rail)

* * *

> So the key contradiction between us is still that you focus on spec but I focus on implementation.

I dunno what to tell you man. 

That's the ethos of my open source project. 

If you don't like it, feel free to fork `gdbstub`. That's what open source is all about.

And for what it's worth, I've recently had the pleasure (displeasure?) to `gdbstub` alongside some non-standard GDB _clients_, at which point, the spec became _extremely_ important, since those alternative clients built their implementation based off the spec doc.

* * *

> Another question: I find the check of `use_implicit_sw_breakpoints` was placed in `run_state_machine`, so `gdbstub` will abort only after it's connected by client. Could we check it beforehand?

The earliest point where I can insert these checks are after the `GdbStub` object has been constructed, which currently also requires an active `Connection` object to be be passed to it during init. I'll mull over if there's some way to lift these checks further up the chain, but I suspect that it might not be possible without some kind of "staggered" `GdbStubBuilder` initialization  (something which I probably won't get around to in time for 0.6).

If this is something you'd really like, I can probably expose some kind of "validate_target" method (somewhere) that can be manually invoked to run these checks, though idk if it's really worth it, given that the checks are moreso a development aid, that shouldn't fire during "steady state" operation of the library.

bet4it:
> the value of `supports_optional_single_step` is independent of the particular mechanism by which single stepping is implemented

I'm very clear of it, but I also know that `supports_optional_single_step` is tied with specific architecture, so what I ask is what `supports_optional_single_step` should be set on MIPS if MIPS works as I described, or you can tell me we don't need to set it on MIPS (so we will use the default `false` value).

* * *

> the fact that optional single stepping was only introduced 2 years after `gdbstub`'s first release is precisely because it's an incredibly niche optional feature, and almost all implementation do end up implementing support for single step! I'd really like to end this back and forth, because we're splitting hairs on a feature that most folks will never care about in the first place.

There are two reasons:

1. If I don't need to implement it, I can just leave it as an empty function if it can't be opted out. It happened when I want to use `gdbstub` to create a minimal dummy example that can be connected by a normal GDB client previously: I only need to implement `read_registers` and `read_addrs`, and leave all other functions empty.
2. You choose ARM program as example and mainly test on it, and nobody else use `gdbstub` on MIPS. If you choose MIPS, you will find it's very necessary: `step` is totally meaningless on MIPS.

> Whether or not it does that on MIPS... I have no idea, I haven't tested it.

I have confirmed it and I really suggest you to test it before `0.6`.

> And to reiterate: this is a guard rail that only really matter when using *off the shelf* `Arch` implementations from `gdbstub_arch`. If you're already implementing your own `Arch`, then you have free reign over whether this guard rail should be enabled or not.

I just don't want to see the situation that you find the name of `supports_optional_single_step` is improper and need to change it after you release `0.6`. Oh, that could be nothing for you🙃 Then you can just do it and ignore me.

* * *
> Dude, that is **literally** the current behavior of the `use_implicit_sw_breakpoints` guard rail!! This is what I've been trying to tell you across countless replies?!

I know it, but how can a new user of `gdbstub` imagine `use_implicit_sw_breakpoints` has relation with single step?

That's what may happen when a new user tries to use `gdbstub` (https://github.com/daniel5151/gdbstub/pull/92#issuecomment-1010011012):

> I just want to ask you, if I'm a normal user of `gdbstub`, and I want to implement `gdbstub` on a MIPS system, how could I know how to implement it? I want to support single step, but after I implement single step, it still can't be used. I don't know what happened, and I find `supports_optional_single_step`, should I change it? And `gdbstub` complains about `use_implicit_sw_breakpoints`, **why do I need to think of breakpoint when I implement single step**?
* * *
> And for what it's worth, I've recently had the pleasure (displeasure?) to gdbstub alongside some non-standard GDB *clients*, at which point, the spec became *extremely* important, since those alternative clients built their implementation based off the spec doc.

Then it's even more improper to add `supports_optional_single_step`. You write `gdbstub` on spec, but if there is a non-standard GDB client that totally follow the spec on this, which means that it will respect the response of `vCont?`, The user of this non-standard GDB client will still face with the annoying `supports_optional_single_step` even if it's meaningless for this non-standard GDB.

And if I write my GDB client that doesn't totally follow the spec and will panic at some point, will you add guard rails for me? Certainly you won't. But why you add a guard rails for GNU GDB which panics because it doesn't follow the spec?
* * *
> If this is something you'd really like, I can probably expose some kind of "validate\_target" method (somewhere) that can be manually invoked to run these checks, though idk if it's really worth it, given that the checks are moreso a development aid, that shouldn't fire during "steady state" operation of the library.

Oh, if it's not easy to change we can just leave it as this.

daniel5151:
> what I ask is what `supports_optional_single_step` should be set on MIPS if MIPS works as I described

Based on what you've told me, it should be set to `false`. If the GDB client has been confirmed to send `s` packets to MIPS targets even in the case that the target itself has not responded with support for `s` support, then this must be set to `false`, as that means this arch does not support optional single step.

It just so happens that the default value is also `false` (i.e: "assume single step is required, until proven that it's possible to leave it optional"), which means you _could_ just leave things as is, but i'd be _nice_ to explicitly document this behavior by opening a PR against the corresponding `gdbstub_arch` implementation.

And like I said, this is not a 0.6 blocker, so if you'd like to open that PR, be my guest.

As part of that PR, please include the protocol logs that show the GDB implementation sending a response to `vCont?` omitting `s` support, and then later on, showing how the GDB client will still send an `s` packet (and how the `gdbstub` proceeds to fail with an unexpected packet).

* * *

> 1. If I don't need to implement it, I can just leave it as an empty function if it can't be opted out.

You're conflating single step _functionality_ and single step _at the protocol level_. If the GDB client request a single step via the GDP RSP, it's my duty as the library author to notify the user of this request - regardless of how they want to handle it.

...unfortunately, the only way to route this message to the user is via the single step IDET + its associated handler, which if left unimplemented, entirely stubs out the packet-parsing codepath related to the `s` packet. This is the intended behavior, and the whole point of using IDETs - you don't pay for what you don't use.

Unfortunately, that means that only notification the user gets in these situations would be an opaque `PacketUnexpected` error, at which point they proceed to open an issue GitHub that I have to respond to.

With this guard rail in place, this error is avoided entirely, since the user is notified that the GDB client will not respect the spec-compliant optional nature of single step for their chosen arch, and it's up to them to implement the single-step IDET (even if it is just a stub implementation).

* * *

> I just don't want to see the situation that you find the name of `supports_optional_single_step` is improper and need to change it after you release `0.6`. Oh, that could be nothing for you🙃 Then you can just do it and ignore me.

I don't entirely understand this statement, but ooookay.

* * *

> I know it, but how can a new user of `gdbstub` imagine `use_implicit_sw_breakpoints` has relation with single step?

I can't keep responding to this question with the same answer. Please see my previous responses and really _think_ about the implications here.

I'm really patient, but responding for a 5th time with the same answer is something I simply cannot do.

* * *

> Then it's even more improper to add `supports_optional_single_step`. You write `gdbstub` on spec, but if there is a non-standard GDB client that totally follow the spec on this, which means that it will respect the response of `vCont?`, The user of this non-standard GDB client will still face with the annoying `supports_optional_single_step` even if it's meaningless for this non-standard GDB.

> But why you add a guard rails for GNU GDB which panics because it doesn't follow the spec?

This is the reality of software engineering in the real world my friend ¯\\\_(ツ)_/¯

Specifically, I know for a fact that 99% of users will be using upstream GDB, which means that any upstream GDB bugs that affect the `gdbstub` experience merit on-by-default work arounds.

In an ideal world, the mainline GDB client would've followed its own spec to-a-tee, the same way it would expect all other implementations to conform to the spec. Alas, either by historical accident, and/or human oversight, it doesn't (and hasn't!) conformed to this edge case of the spec in a looooong time, which means that as the author of `gdbstub`, I'm _choosing_ to work around the current issues present in the GDB client in my library. Not doing so would result in my bug reports filed against _my project_, and I'd like to nip those bug reports in the bud pre-emptively.

That said, if you are one of the rare folks using `gdbstub` with a non-standard GDB client, I'm offering the _flexibility_ to disable these guard rails yourself, with the promise that You Know What You're Doing ™️ 

Again, to be perfectly crystal clear: these are _guard rails_, not _maximum security holding cells_! 
If you don't care this guard rail (or any other one for that matter), just turn them off!

daniel5151:
Anyways, I'm not sure if I have the time or energy to keep going on in this back and forth.

I'm happy to read + ack any response you might have, but I'm pretty convinced that this functionality should remain in `gdbstub` and ship as part of 0.6.

If you don't like it, you can just turn it off (or in your case, for MIPS, keep the default behavior, because it sounds like you'll need to support single step packets anyways).

And to be clear: these guard rails are a new idea I'm testing out as part of the 0.6 release. If it turns out folks categorically _don't_ like them, then I might remove them at some point in the future. 

I'm not claiming to know what's perfect - I'm only human.

Cheers

bet4it:
> [5bc5831](https://github.com/daniel5151/gdbstub/commit/5bc5831b61a5e0dfe23ce548e790cd6c4d4c6888) + [1c6560e](https://github.com/daniel5151/gdbstub/commit/1c6560e7082d9c7d37c52bf15d01397ebf4a0fdf) has made all forms of resume optional tada

Just found that `support_single_step` is a part of `SingleThreadResume`? Can't I only support single step but not continue?
And as I said in #59, we need to rename `SingleThreadResume` to `SingleThreadContinue`?

daniel5151:
Skimming through #59, I remember why I went with this approach. It's a two-parter:

Pragmatically:

In part due to the fact that `gdbstub` uses multi-process extensions under the hood, `vCont` responses will pretty much always include a `-1;c` component (i.e: resume all other threads), even in the case of attempting to single-step the target. In order to parse + handle this aspect of the `vCont` packet, the target needs to implement the continue IDET so I can properly report this condition to the target.

Ideally:

The spec states the following:

> Stubs that only control single-threaded targets can implement run control with the ‘c’ (continue) command, and if the target architecture supports hardware-assisted single-stepping, the ‘s’ (step) command. Stubs that support multi-threading targets should support the ‘vCont’ command.

Admittedly, this is a bit ambiguous, but I comprehend this statement as: IF you support resuming the target, you MUST support continue, and OPTIONALLY support single-stepping (IF the target support hardware-assisted single-stepping).

* * *

FWIW, I can understand why single step and continue _should_ be orthogonal features, but the reality is that they seem to be inexorably intertwined with one another. I can see a future where I end up circling back to this issue and making the two orthogonal within `gdbstub`, but doing so would require some refactoring + adding a new guard rail to work around the aforementioned pragmatic points.

This is a niche enough edge case that I'm not going to be blocking 0.6 to change this, but if you feel strongly about this, consider opening a dedicated tracking issue for this.

* * *

Oh, and lastly - if you _really_ only want to support single step without continue, you can just ignore continue requests (or treat them as single-step requests). Not ideal, but I think the GDB client should be resilient enough to handle that.

daniel5151:
Quite late to acknowledge this, but it seems there's actually been some movement on https://sourceware.org/bugzilla/show_bug.cgi?id=28440 (re: how certain platforms don't respect optional single stepping).

I haven't tested it, but things _might_ finally work properly on GDB main? Would be good to check and see if we can update certain docs.

daniel5151:
Ouch, didn't realize that patch never got merged. That's unfortunate.

I saw you sent in https://github.com/daniel5151/gdb-optional-step-bug/pull/5 which updates, modernizes, and extends the repro repo - thanks for that. I'll take a look at that PR shortly.

You should definitely set up a bugzilla account and leave a comment on that bug, so folks are aware of its status in 2025 (and linking to any new context / findings you've dug up)

PR #93, opened 2021-11-04
Fix missing import for trace-pkt.

gz:


daniel5151:
oops, good catch!

gz:
I finally got around to updating to dev/0.6 again. Everything still seems to work. Next, I'll write some more integration tests to see if the Ctrl+C issue got resolved as well with the new state machine.

daniel5151:
Glad to hear! Let me know once it's confirmed working (by commenting on the appropriate issue), and we can close it out :)

gz:
One thing I'm not sure about: If I'm rolling my own state-machine code and have to call `interrupt_handled` for a Ctrl+C what kind of stop reason should I report (https://github.com/daniel5151/gdbstub/blob/f787abf83e050970d63d10ad87dc9fe93a45ecf6/src/gdbstub_impl/mod.rs#L571).

It seems the no-std example passes in None, but that seems wrong judging the doc comment of the function?

daniel5151:
We really should move this discussion over to #89...

I probably should've told you to look at the _users_ of run_blocking as well, as that makes it a bit more clear. e.g:
https://github.com/daniel5151/gdbstub/blob/f787abf83e050970d63d10ad87dc9fe93a45ecf6/examples/armv4t/main.rs#L127-L139

Issue #53, opened 2021-05-06
Runtime-configurable `Arch` selection


daniel5151:
Extracted from a back-and-forth on https://github.com/daniel5151/gdbstub/issues/31#issuecomment-795811837

* * *

With the current `Arch` architecture, it's actually pretty easy to "punch through" the type-safe abstractions by implementing an `Arch` that looks something like this:

```rust
// not actually tested, but something like this ought to work

pub enum PassthroughArch {}

impl Arch for PassthroughArch {
    type Usize = u64; // something reasonably big
    type Registers = PassthroughRegs;
    type RegId = PassthroughRegId;
    ...
}

pub struct PassthroughRegs {
    len: usize,
    buffer: [u8; MAX_SIZE], // a vec could also work
}

// push logic up a level into `Target`
impl Registers for PassthroughRegs {
    fn gdb_serialize(&self, mut write_byte: impl FnMut(Option<u8>)) {
        for byte in &self.buffer[..self.len] { write_byte(Some(byte)) }
    }

    fn gdb_deserialize(&mut self, bytes: &[u8]) -> Result<(), ()> {
        self.buffer[..bytes.len()].copy_from_slice(bytes)
    }
}

pub struct PassthroughRegId(usize);
impl RegId for PassthroughRegId { ... } // elided for brevity, should be straightforward

impl Target {
    type Arch = PassthroughArch;

    // .... later on, in the appropriate IDET ... //

    fn read_registers(
        &mut self, 
        regs: &mut PassthroughRegs<YourTargetSpecificMetadata>
    ) -> TargetResult<(), Self> {
        // write data directly into `regs.buffer` + `regs.len`
        if self.cpu.in_foo_mode() { /* one way */  } else { /* other way */ }
        // can re-use existing `impl Registers` types from `gdbstub_arch`, 
        // calling `gdb_serialize` and `gdb_deserialize` directly...
        if using_arm {
            let mut regs = gdbstub_arch::arm::regs::ArmCoreRegs::default();
            regs.pc = self.pc;
            // ...
            let mut i = 0;
            regs.gdb_serialize(|b| { regs.buffer[i] = b.unwrap_or(0); i++; })
        }
    }

    // similarly for `write_registers`
}
```

This can then be combined with the recently added `TargetDescriptionXmlOverride` IDET to effectively create a runtime configurable `Arch` + `Target` (see #43)

While this will work in `gdbstub` _right now_, this approach does have a few downsides:

[...]

PR #95, opened 2022-01-25
Update `single_step_gdb_behavior` of all architectures with proper value

bet4it:
### Description

I create a generic stub [`gdb_generic.rs`](https://github.com/bet4it/gdb-optional-step-bug/blob/dac95d5e17ec17603c0802c384b81e699fc29d47/src/gdb_generic.rs) which can be used to test the behavior of single step of all architectures. All you need is `set architecture` with the architecture you want to test before `target remote`, for example:
```
(gdb) set architecture powerpc:common
The target architecture is set to "powerpc:common".
(gdb) target remote :9001
Remote debugging using :9001
warning: No executable has been specified and target does not support
determining executable automatically.  Try using the "file" command.
0x00000000 in ?? ()
(gdb) stepi
0x00000000 in ?? ()
```
I don't consider the situation that `single_step_gdb_behavior` should be set to `Optional` because as I said in https://github.com/daniel5151/gdbstub/pull/92#issuecomment-971641019, it only possible happens on arm.

This generic stub uses the Dynamic Arch selection (#53) way.
And you may notice that `resume` and `write_registers` are totally useless for this real program, but I must implement them. We may make them optional?

### API Stability

### Checklist

- Implementation
  - [x] `cargo fmt` was run
  - [x] All tests pass
- Documentation
  - [ ] rustdoc + approprate inline code comments
  - [ ] Updated CHANGELOG.md
  - [x] I have tested this code in my project, and to the best of my knowledge, it is working as intended.

### DIFF
--- a/src/arch.rs
+++ b/src/arch.rs
@@ -250,4 +250,8 @@ pub enum SingleStepGdbBehavior {
     ///
     /// e.g: MIPS
     Ignored,
+    /// Unknown behavior - no one has tested this platform yet. If possible,
+    /// please conduct a test + upstream your findings to `gdbstub_arch`.
+    #[doc(hidden)]
+    Unknown,
 }
--- a/src/arch.rs
+++ b/src/arch.rs
@@ -202,19 +202,6 @@ pub trait Arch {
     /// pre-init error that informs the user of the potential issues they may
     /// run into.
     ///
-    /// # `Unknown` implementations
-    ///
-    /// Because this method was only introduced in `gdbstub` version 0.6, there
-    /// are many existing `Arch` implementations in the
-    /// [`gdbstub_arch`](https://docs.rs/gdbstub_arch/) companion crate that
-    /// have not yet been tested and updated what kind of behavior they exhibit.
-    ///
-    /// These implementations currently return
-    /// [`SingleStepGdbBehavior::Unknown`], which will result in a pre-init
-    /// error that notifies users of this issue, along with imploring them
-    /// to be a Good Citizen and discover + upstream a proper implementation
-    /// of this method for their `Arch`.
-    ///
     /// # Writing a proper implementation
     ///
     /// To check whether or not a particular architecture exhibits this
@@ -263,7 +250,4 @@ pub enum SingleStepGdbBehavior {
     ///
     /// e.g: MIPS
     Ignored,
-    /// Unknown behavior - no one has tested this platform yet. If possible,
-    /// please conduct a test + upstream your findings to `gdbstub_arch`.
-    Unknown,
 }
--- a/gdbstub_arch/src/msp430/mod.rs
+++ b/gdbstub_arch/src/msp430/mod.rs
@@ -19,8 +19,7 @@ impl Arch for Msp430 {
 
     #[inline(always)]
     fn single_step_gdb_behavior() -> SingleStepGdbBehavior {
-        // TODO: update with proper value
-        SingleStepGdbBehavior::Unknown
+        SingleStepGdbBehavior::Required
     }
 }
 
@@ -39,7 +38,6 @@ impl Arch for Msp430X {
 
     #[inline(always)]
     fn single_step_gdb_behavior() -> SingleStepGdbBehavior {
-        // TODO: update with proper value
-        SingleStepGdbBehavior::Unknown
+        SingleStepGdbBehavior::Required
     }
 }
--- a/gdbstub_arch/src/ppc/mod.rs
+++ b/gdbstub_arch/src/ppc/mod.rs
@@ -27,7 +27,6 @@ impl<RegIdImpl: RegId> Arch for PowerPcAltivec32<RegIdImpl> {
 
     #[inline(always)]
     fn single_step_gdb_behavior() -> SingleStepGdbBehavior {
-        // TODO: update with proper value
-        SingleStepGdbBehavior::Unknown
+        SingleStepGdbBehavior::Required
     }
 }
--- a/gdbstub_arch/src/riscv/mod.rs
+++ b/gdbstub_arch/src/riscv/mod.rs
@@ -19,13 +19,12 @@ impl Arch for Riscv32 {
     type BreakpointKind = usize;
 
     fn target_description_xml() -> Option<&'static str> {
-        Some(r#"<target version="1.0"><architecture>riscv</architecture></target>"#)
+        Some(r#"<target version="1.0"><architecture>riscv:rv32</architecture></target>"#)
     }
 
     #[inline(always)]
     fn single_step_gdb_behavior() -> SingleStepGdbBehavior {
-        // TODO: update with proper value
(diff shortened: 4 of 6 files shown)


src/arch.rs
```diff
@@ -263,7 +250,4 @@ pub enum SingleStepGdbBehavior {
     ///
     /// e.g: MIPS
     Ignored,
-    /// Unknown behavior - no one has tested this platform yet. If possible,
-    /// please conduct a test + upstream your findings to `gdbstub_arch`.
-    Unknown,
```
daniel5151, line 268: Removing this variant would technically be a breaking API change, and given how "small" the change is, I'd rather not push out a whole new gdbstub 0.7 for that.

Instead, lets just mark this variant as `#[doc(hidden)]`, and then I'll spin up a tracking task to remove the variant entirely whenever the next breaking release comes out.

daniel5151 (changes requested):

daniel5151:
> I don't consider the situation that `single_step_gdb_behavior` should be set to `Optional` because as I said in [#92 (comment)](https://github.com/daniel5151/gdbstub/pull/92#issuecomment-971641019), it only possible happens on arm.

Ehh, I don't have the energy to argue here. Given that going from `Required` -> `Optional` is a "loosening" of behavior, I don't mind if it ends up getting changed as some point down the line.

* * *

> And you may notice that `resume` and `write_registers` are totally useless for this real program, but I must implement them. We may make them optional?

wrt. `resume`: I totally forgot to rename `resume` to `cont` for the single thread case, that's my bad. That said, I do consider support for the `c` resume reason a requirement for any "resumable" target, hence why the method is non-optional. We've already had this discussion, and I think we came to the conclusion of "agree to disagree" here.

wrt. `write_registers`: once again, please refer to the following line in the GDB RSP spec:

> At a minimum, a stub is required to support the ‘?’ command to tell GDB the reason for halting, ‘g’ and ‘G’ commands for register access, and the ‘m’ and ‘M’ commands for memory access.

src/arch.rs
bet4it: Why you need to release `gdbstub` 0.7 only for this?
You can just leave this commit in the `dev/0.7` branch and release it along with all the other changes when they are done in `gdbstub` 0.7.
And you can release `gdbstub_arch` 0.2.1 for this commit. Then people won't encounter breaking API change for this even `gdbstub` 0.7 is released in the future.

bet4it: OK, I add `SingleStepGdbBehavior::Unknown` back, although I'm really reluctant to do this.
I think it's really not necessary to push out `gdbstub` 0.6.1 that the only difference with 0.6.0 is `#[doc(hidden)]`, because even if `Unknown` exists, nobody will use it. People won't notice it if they use the defined architecture, and if they want to create a new architecture, they will clear it is not something they need. You only need to push out `gdbstub_arch` 0.2.1.

> That said, given that these changes would all be non-breaking and land in `gdbstub_arch`, it's fine to have some `Unknown`s for now and update things as folks have time to look into each arch.

So the first commit after `gdbstub` 0.6.0 is a breaking API change🙃
* * *
Fine, it's your project and you are always right. You can just revert the later commit after you create the `dev/0.7` branch🙂

daniel5151: When I implemented the Unknown variant, I knew full well that it'd end up getting removed fairly quickly, and that I'd need to push out a new minor `gdbstub_arch` release, along with (possibly) a new minor `gdbstub` release that soft-deprecates the newly added variant.

While it may seem silly to add a feature to 0.6 that immediately gets soft deprecated, the reality is that - for my own sanity - I really just wanted to get 0.6 out the door ¯\\\_(ツ)_/¯

> So the first commit after `gdbstub` 0.6.0 is a breaking API change🙃

Not super sure what you mean by that.

Of course, whatever first commit ends up in `dev/0.7` will be a breaking change. That's the whole point of opening a new `dev` branch - to put in any new breaking changes.

It could have just as easily been the case that you disappeared for a few months, and in the meantime, someone else came along and opened a PR to add a new non-breaking protocol extension to the library, which wouldn't even require opening a new `dev/0.7` branch

daniel5151 (approved):

daniel5151:
`gdbstub_arch` 0.2.1 has been published to crates.io that includes these changes :)

I've decided to hold off on releasing `gdbstub` 0.6.1 just yet, since I'm inclined to agree that a whole new minor commit just for a doc change seems a bit silly. 

If I misunderstood your intent, and you'd like to see 0.6.1 published with this doc tweak, do let me know.

bet4it:
> I've decided to hold off on releasing `gdbstub` 0.6.1 just yet, since I'm inclined to agree that a whole new minor commit just for a doc change seems a bit silly.

Yes, that's what I want!
But if we don't make a new release for this, why I need to add `SingleStepGdbBehavior::Unknown` back🙃 That's why I said I'm reluctant to do this.
If you still have some big non-breaking changes that want to be included in `0.6`, then it's understandable.

daniel5151:
I should probably clarify how I structure branches on this repo:

- `master` only ever includes semver-compatible changes to latest release of `gdbstub` (which is currently 0.6)
- `dev/0.X` branches are `master` + any breaking changes

Do you see why I asked you to re-add `Unknown`?

This PR is being merged into `master`, which means it must be semver compatible with 0.6, which means we cannot remove an enum variant as part of this PR. That would be a semver violation.

Instead, we've merged this PR into `master` in a semver-compatible way (i.e: marking `Unknown` as `doc(hidden)`, and I've opened a tracking task to remove the variant entirely whenever I end up creating a new `dev/0.7` branch for breaking changes.

bet4it:
IIRC, you created `dev/0.6` branch as soon as you release `0.5`, and didn't maintain a master branch that kept compatibility with `0.5` at that time, that's what I want this time: we can just do our work on `dev/0.7` branch if you don't have further non-breaking improvements on `0.6`.

daniel5151:
Yeah, that was just a fluke :)
It just so happened that after releasing 0.5, I very quickly realized that there were some breaking changes I needed to make. I was hoping to release at least a couple minor releases, but that never really happened...

The 0.4 -> 0.5 transition is a better example - in that case, there were quite a few non-breaking features implemented on `master` before I ended up making `dev/0.5` for breaking changes.

* * *

Just so you know, now that 0.6 is released, I'm going to be taking a step back from `gdbstub` for a bit. Of course, I'll continue reviewing PRs and addressing issues, but I won't be doing any active feature development for a while.

I'm hoping that in that time, folks will still send smaller, semver-compatible protocol extension IDET PRs (targeting the `master` branch), and that I could release a couple 0.6.x minor versions with those changes.

That said, it sounds like you have some breaking-change PRs you'd like to send my way. If that's the case, I've gone ahead and make a `dev/0.7` branch, which you can target those PRs against :)

bet4it:
😅Oh, I don't have other breaking-change PRs to be sent, this is the only breaking-change PR😅

PR #98, opened 2022-02-19
do not fail on unknown HostIoOpenFlags

mrk-its:
### Description

gdbstub fails when unknown file open flags are passed with:
```
gdbstub encountered a fatal error: Could not parse the packet into a valid command: MalformedCommand
```

This bug appeared with lldb sending `platform put-file` command, appending following `File:open` arguments:
`612e6f7574,40000601,000001ed`

The simple solution is to use `from_bits_truncate` to ignore unknown flags.

### API Stability

### Checklist

- Implementation
  - [x] `cargo fmt` was run
  - [x] All tests pass
 
- Documentation
  - [x] Updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@ All notable changes to this project will be documented in this file.
 
 This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+# 0.6.1
+
+- ignore unknown vFile:open flags
+
 # 0.6.0
 
 After over a half-year of development, `gdbstub` 0.6 has finally been released!
--- a/src/protocol/commands/_vFile_open.rs
+++ b/src/protocol/commands/_vFile_open.rs
@@ -20,8 +20,8 @@ impl<'a> ParseCommand<'a> for vFileOpen<'a> {
             [b':', body @ ..] => {
                 let mut body = body.splitn_mut_no_panic(3, |b| *b == b',');
                 let filename = decode_hex_buf(body.next()?).ok()?;
-                let flags = HostIoOpenFlags::from_bits(decode_hex(body.next()?).ok()?)?;
-                let mode = HostIoOpenMode::from_bits(decode_hex(body.next()?).ok()?)?;
+                let flags = HostIoOpenFlags::from_bits_truncate(decode_hex(body.next()?).ok()?);
+                let mode = HostIoOpenMode::from_bits_truncate(decode_hex(body.next()?).ok()?);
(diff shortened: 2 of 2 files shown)


daniel5151:
While this PR does solve your immediate issue, it is not a proper solution to the problem. 
Namely, it seems that the LLDB RSP deviates from GDB RSP wrt how it handles vFile packets.

See: https://github.com/llvm/llvm-project/blob/ec642ceebc1aacc8b16249df7734b8cf90ae2963/lldb/docs/lldb-platform-packets.txt#L6-L11

I'll try to find some time soon to think of the best way to work around this issue, but unfortunately, I cannot merge this PR in its current state.

Note that as it stands, `gdbstub` doesn't explicitly support `lldb`, so this isn't strictly a "bug" as it is a new feature request (i.e: supporting LLDB).

mrk-its:
np, so I'll stay with my fork

daniel5151:
I did a bit more digging and found this:

https://github.com/llvm/llvm-project/blob/ec642ceebc1aacc8b16249df7734b8cf90ae2963/lldb/include/lldb/Host/File.h#L47-L66

Look like LLVM's flags are a strict superset of GDB flags, which means it should be possible to support these additional flags without breaking existing GDB implementation / requiring any breaking API changes.

If you'd like to send a PR that adds these additional variants, I'd be happy to merge that one.

daniel5151:
oh, wait, maybe not actually...

looks like this was only a recent change, and older versions of lldb will use different enum values: https://github.com/llvm/llvm-project/commit/8bbef4f9afd8b5f6b4435d5bab48b75f048f353c

I suspect we'd want to support older clients as well, which means this continues to be a tricky problem to solve nicely...

mrk-its:
> oh, wait, maybe not actually...
> 
> looks like this was only a recent change, and older versions of lldb will use different enum values: [llvm/llvm-project@8bbef4f](https://github.com/llvm/llvm-project/commit/8bbef4f9afd8b5f6b4435d5bab48b75f048f353c)
> 
> I suspect we'd want to support older clients as well, which means this continues to be a tricky problem to solve nicely...

Currently `gdbstub` fails for both old and new versions, so adding support for new ones would be a big improvement :)

daniel5151:
Yeah, you know what - fair enough! We can punt the hard problem of supporting older LLDB clients down the line, and get you unblocked for now in a non-breaking way.

So sure, feel free to send a PR my way with these new variants :)

daniel5151:
FYI, I've opened a tracking issue for the broader LLDB compatibility story over at #99

PR #102, opened 2022-04-24
Update real-world examples

bet4it:
Hi Daniel,

After a long delay, I finally release my project [udbserver](https://github.com/bet4it/udbserver) which heavily bases on `gdbstub`. Thanks a lot, you save me a lot of time and make the implementation very clear!

I'm a newbie to Rust and this is my first Rust project. Really hope for your advice on it to help me improve it. For example, how to reduce the use of `unsafe`, and is this possible to implement Ctrl-C in my project.

Best wishes!

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,20 +125,22 @@ When using `gdbstub` in `#![no_std]` contexts, make sure to set `default-feature
 ### Real-World Examples
 
 -   Virtual Machine Monitors (VMMs)
-    -   [crosvm](https://google.github.io/crosvm/running_crosvm/usage.html#gdb-support) - The Chrome OS Virtual Machine Monitor (x64)
-    -   [Firecracker](https://firecracker-microvm.github.io/) - A lightweight VMM developed by AWS - feature is in [PR](https://github.com/firecracker-microvm/firecracker/pull/2333)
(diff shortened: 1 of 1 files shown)


daniel5151:
Heyo, I manually went ahead and updated the real-world examples section of the readme, adding udbserver (along with cleaning up some of the other entries, slightly differently than how you did it in this PR)

daniel5151:
Some thoughts on `udbserver`:

While I likely can't offer a full code review anytime soon, I took a brief skim through the code, and I've got some preliminary thoughts:

- It seems that a bulk of your `unsafe` usage comes from using global `mut static` variables to hold on to state. Instead, you should consider using something like a `static mut STATE: OnceCell<Mutex<State>> = ...` to synchronize access to this state. Alternatively, you can avoid using global state alltogether, and package your state into a `struct` which the resulting caller has to manage the lifetime of.
- As for supporting ctrl-c... I suspect that's something you'll need to dig into yourself, as supporting it cleanly typically requires intimate knowledge of the overall architecture of the application `gdbstub` is being integrated into, which I simply don't have. I would suggest looking at some of examples (both in-tree and real-world) for how different folks have done it.

Issue #99, opened 2022-02-20
LLDB Compatibility


daniel5151:
This is a meta-issue to discuss + track `gdbstub`'s LLDB compatibility story.

### Overview 

LLDB uses the GDB Remote Serial Protocol for remote debugging, and for the most part, LLDB works "out of the box" with GDB server implementations. Unfortunately, there are some differences between how LLDB and GDB implement the Remote Serial Protocol, resulting in potential issues when mixing and matching servers/clients.

In some cases, LLDB has added new protocol extensions that are disjoint from (and therefore backwards compatible with) the base GDB protocol. The de-facto documentation for these LLDB specific extensions can be found here:
- https://github.com/llvm/llvm-project/blob/main/lldb/docs/lldb-gdb-remote.txt
- https://github.com/llvm/llvm-project/blob/main/lldb/docs/lldb-platform-packets.txt

Unfortunately, there are some cases where LLDB happened to implement certain packets _differently_ from GDB, resulting in breaking incompatibilities. The most pertinent example of this is wrt how LLDB handles `vFile` packets. This blog post does a good job providing an overview of the various LLDB-GDB RSP compatibility hazards: https://www.moritz.systems/blog/improving-gdb-protocol-compatibility-in-lldb/

Moreover, as mentioned in the blog post above, LLDB has actually made _breaking changes_ to its RSP between versions! Unlike GDB, using a newer LLDB client with an older LLDB server is _not_ supported!

### Supporting LLDB clients in `gdbstub`

With these challenges in mind, what would be the best way to support LLDB clients in `gdbstub`?

For the most part, it should be pretty straightforward to support arbitrary LLDB extensions in `gdbstub`. After all, adding support for any particular LLDB packet should be no different from adding support for a new GDB packet. 

Indeed, the bulk of the complexity in supporting LLDB clients is figuring out how to handle the _inconsistencies_ between the LLDB and GDB RSPs.

- Should `gdbstub` include some kind of configuration to target a specific LLDB protocol version?
- Would it be feasible to somehow support _all_ LLDB clients via some kind of version-detection mechanism?
- Should LLDB extensions "bleed into" the existing APIs, or should LLDB extensions be clearly marked as "separate" from the main GDB protocol?
  - e.g: should the existing vFile:open API be updated to include LLDB-specific vFile:open flags?
[...]

PR #100, opened 2022-02-21
add lldb-specific HostIoOpenFlags

mrk-its:
### Description

This PR adds lldb-specific HostIoOpenFlags, as proposed in #99

### API Stability

- [x] This PR does not require a breaking API change

### Checklist

- Implementation
  - [x] `cargo build` compiles without `errors` or `warnings`
  - [x] `cargo clippy` runs without `errors` or `warnings`
  - [x] `cargo fmt` was run
  - [x] All tests pass
- Documentation
  - [x] rustdoc + approprate inline code comments
  - [x] Updated CHANGELOG.md

PR #66, opened 2021-07-30
Support Host I/O operations

bet4it:
### Description

Support Host I/O operations 

Refer: #32 

### API Stability

- [x] This PR does not require a breaking API change

### Checklist

- Implementation
  - [x] `cargo build` compiles without `errors` or `warnings`
  - [x] `cargo clippy` runs without `errors` or `warnings`
  - [x] `cargo fmt` was run
  - [x] All tests pass
- Documentation
  - [x] rustdoc + approprate inline code comments
  - [ ] Updated CHANGELOG.md
  - [x] (if appropriate) Added feature to "Debugging Features" in README.md
- _If implementing a new protocol extension IDET_
  - [x] Included a basic sample implementation in `examples/armv4t`
  - [x] Included output of running `examples/armv4t` with `RUST_LOG=trace` + any relevant GDB output under the "Validation" section below
  - [x] Confirmed that IDET can be optimized away (using `./scripts/test_dead_code_elim.sh` and/or `./example_no_std/check_size.sh`)
  - [ ] **OR** Implementation requires adding non-optional binary bloat (please elaborate under "Description")
- _If upstreaming an `Arch` implementation_
  - [ ] I have tested this code in my project, and to the best of my knowledge, it is working as intended.

### Validation

<details>
<summary>GDB output</summary>

```
(gdb) remote get /tmp/remote.txt /tmp/local.txt
Successfully fetched file "/tmp/remote.txt".
(gdb) remote put /tmp/local.txt /tmp/remote.txt
Successfully sent file "/tmp/local.txt".
(gdb) remote delete /tmp/remote.txt
Successfully deleted file "/tmp/remote.txt"
```

</details>

<details>
<summary>armv4t output</summary>

```
    Finished dev [unoptimized + debuginfo] target(s) in 0.02s
     Running `target/debug/examples/armv4t`
loading section ".text" into memory from [0x55550000..0x55550078]
Setting PC to 0x55550000
Waiting for a GDB connection on "127.0.0.1:9001"...
Debugger connected from 127.0.0.1:36992
 TRACE gdbstub::protocol::recv_packet > <-- +
 TRACE gdbstub::protocol::recv_packet > <-- $qSupported:multiprocess+;swbreak+;hwbreak+;qRelocInsn+;fork-events+;vfork-events+;exec-events+;vContSupported+;QThreadEvents+;no-resumed+;xmlRegisters=i386#6a
[...]

PR #103, opened 2022-05-31
add qRegisterInfo packet support

jawilk:
### Description

This implements the LLDB specifc `qRegisterInfo` packet (documentation [here](https://github.com/llvm/llvm-project/blob/50f2e49924566330ea9aa731908f2864603bf4fb/lldb/docs/lldb-gdb-remote.txt#L583)).

Relates to #99 
I was working on a target where I had some trouble getting the .xml target description formatting right so I opted to use the `qRegisterInfo` packet instead which pretty much worked right away. I haven't looked into unifying them as mentioned in the compatibility issue as of now though.
One caveat is that some values in the reply need to be in decimal representation. Due to this I added a `write_dec()` method in the `ResponseWriter` (I'm not sure this was actually necessary). Since this packet is only used by `LLDB` it might be reasonable to just hide it behind a feature flag. Besides that, if enabled, the `Register` struct in `gdbstub/src/target/ext/register_info.rs` has a few `&'static str` members which lead to a bigger binary as well.

### API Stability

### Checklist

- Implementation
  - [x] `cargo fmt` was run
  - [x] All tests pass
- Documentation
  - [x] Updated CHANGELOG.md
  - [x] Included a basic sample implementation in `examples/armv4t`
    ->Done but might be better to comment it out because it only applies to `LLDB`
    ->attached the LLDB output instead
  - [x] Confirmed that IDET can be optimized away (using `./scripts/test_dead_code_elim.sh` and/or `./example_no_std/check_size.sh`)
  - [x] **OR** Implementation requires adding non-optional binary bloat (please elaborate under "Description")
    ->Additional `write_dec()` method

### Validation

<details>
<summary>LLDB output</summary>

```
(lldb) gdb-remote 9001
lldb-10          <   1> send packet: +
lldb-10          history[1] tid=0x769a <   1> send packet: +
lldb-10          <  19> send packet: $QStartNoAckMode#b0
lldb-10          <   1> read packet: +
lldb-10          <   6> read packet: $OK#9a
lldb-10          <   1> send packet: +
lldb-10          <  45> send packet: $qSupported:xmlRegisters=i386,arm,mips,arc#74
lldb-10          < 333> read packet: $PacketSize=1000;vContSupported+;multiprocess+;QStartNoAckMode+;ReverseContinue+;ReverseStep+;QDisableRandomization+;QEnvironmentHexEncoded+;QEnvironmentUnset+;QEnvironmentReset+;QStartupWithShell+;QSetWorkingDir+;swbreak+;hwbreak+;QCatchSyscalls+;qXfer:features:read+;qXfer:memory-map:read+;qXfer:exec-file:read+;qXfer:auxv:read+#fa
lldb-10          <  26> send packet: $QThreadSuffixSupported#e4
lldb-10          <   4> read packet: $#00
lldb-10          <  27> send packet: $QListThreadsInStopReply#21
lldb-10          <   4> read packet: $#00
lldb-10          <  13> send packet: $qHostInfo#9b
lldb-10          <   4> read packet: $#00
lldb-10          <  10> send packet: $vCont?#49
lldb-10          <  19> read packet: $vCont;c;C;s;S;r#0f
lldb-10          <  27> send packet: $qVAttachOrWaitSupported#38
lldb-10          <   4> read packet: $#00
lldb-10          <  23> send packet: $QEnableErrorStrings#8c
lldb-10          <   4> read packet: $#00
lldb-10          <  16> send packet: $qProcessInfo#dc
lldb-10          <   4> read packet: $#00
lldb-10          <   6> send packet: $qC#b4
lldb-10          <   4> read packet: $#00
lldb-10          <  16> send packet: $qfThreadInfo#bb
lldb-10          <   7> read packet: $m01#ce
lldb-10          <  16> send packet: $qsThreadInfo#c8
lldb-10          <   5> read packet: $l#6c
lldb-10          <   5> send packet: $?#3f
lldb-10          <   7> read packet: $S05#b8
lldb-10          <  16> send packet: $qProcessInfo#dc
lldb-10          <   4> read packet: $#00
lldb-10          <  18> send packet: $qRegisterInfo0#72
lldb-10          < 107> read packet: $name:r0;alt-name:r0;bitsize:32;offset:0;encoding:uint;format:hex;set:General Purpose Registers;dwarf:0;#bb
lldb-10          <  18> send packet: $qRegisterInfo1#73
lldb-10          < 107> read packet: $name:r1;alt-name:r1;bitsize:32;offset:4;encoding:uint;format:hex;set:General Purpose Registers;dwarf:1;#c2
lldb-10          <  18> send packet: $qRegisterInfo2#74
lldb-10          < 107> read packet: $name:r2;alt-name:r2;bitsize:32;offset:8;encoding:uint;format:hex;set:General Purpose Registers;dwarf:2;#c9
lldb-10          <  18> send packet: $qRegisterInfo3#75
lldb-10          < 108> read packet: $name:r3;alt-name:r3;bitsize:32;offset:12;encoding:uint;format:hex;set:General Purpose Registers;dwarf:3;#f7
lldb-10          <  18> send packet: $qRegisterInfo4#76
lldb-10          < 108> read packet: $name:r4;alt-name:r4;bitsize:32;offset:16;encoding:uint;format:hex;set:General Purpose Registers;dwarf:4;#fe
lldb-10          <  18> send packet: $qRegisterInfo5#77
lldb-10          < 108> read packet: $name:r5;alt-name:r5;bitsize:32;offset:20;encoding:uint;format:hex;set:General Purpose Registers;dwarf:5;#fc
lldb-10          <  18> send packet: $qRegisterInfo6#78
lldb-10          < 108> read packet: $name:r6;alt-name:r6;bitsize:32;offset:24;encoding:uint;format:hex;set:General Purpose Registers;dwarf:6;#03
lldb-10          <  18> send packet: $qRegisterInfo7#79
lldb-10          < 108> read packet: $name:r7;alt-name:r7;bitsize:32;offset:28;encoding:uint;format:hex;set:General Purpose Registers;dwarf:7;#0a
lldb-10          <  18> send packet: $qRegisterInfo8#7a
lldb-10          < 108> read packet: $name:r8;alt-name:r8;bitsize:32;offset:32;encoding:uint;format:hex;set:General Purpose Registers;dwarf:8;#08
lldb-10          <  18> send packet: $qRegisterInfo9#7b
lldb-10          < 108> read packet: $name:r9;alt-name:r9;bitsize:32;offset:36;encoding:uint;format:hex;set:General Purpose Registers;dwarf:9;#0f
lldb-10          <  18> send packet: $qRegisterInfoa#a3
lldb-10          < 111> read packet: $name:r10;alt-name:r10;bitsize:32;offset:40;encoding:uint;format:hex;set:General Purpose Registers;dwarf:10;#82
lldb-10          <  18> send packet: $qRegisterInfob#a4
lldb-10          < 111> read packet: $name:r11;alt-name:r11;bitsize:32;offset:44;encoding:uint;format:hex;set:General Purpose Registers;dwarf:11;#89
lldb-10          <  18> send packet: $qRegisterInfoc#a5
lldb-10          < 111> read packet: $name:r12;alt-name:r12;bitsize:32;offset:48;encoding:uint;format:hex;set:General Purpose Registers;dwarf:12;#90
lldb-10          <  18> send packet: $qRegisterInfod#a6
lldb-10          < 120> read packet: $name:sp;alt-name:sp;bitsize:32;offset:52;encoding:uint;format:hex;set:General Purpose Registers;dwarf:13;generic:sp;#dd
lldb-10          <  18> send packet: $qRegisterInfoe#a7
lldb-10          < 109> read packet: $name:lr;alt-name:lr;bitsize:32;offset:56;encoding:uint;format:hex;set:General Purpose Registers;dwarf:14;#a3
lldb-10          <  18> send packet: $qRegisterInfof#a8
lldb-10          < 120> read packet: $name:pc;alt-name:pc;bitsize:32;offset:60;encoding:uint;format:hex;set:General Purpose Registers;dwarf:15;generic:pc;#ae
lldb-10          <  19> send packet: $qRegisterInfo10#a3
lldb-10          < 129> read packet: $name:padding;alt-name:padding;bitsize:32;offset:64;encoding:vector;format:vector-uint8;set:Floating Point Registers;dwarf:16;#6e
lldb-10          <  19> send packet: $qRegisterInfo11#a4
lldb-10          < 129> read packet: $name:padding;alt-name:padding;bitsize:32;offset:68;encoding:vector;format:vector-uint8;set:Floating Point Registers;dwarf:17;#73
lldb-10          <  19> send packet: $qRegisterInfo12#a5
lldb-10          < 129> read packet: $name:padding;alt-name:padding;bitsize:32;offset:72;encoding:vector;format:vector-uint8;set:Floating Point Registers;dwarf:18;#6f
lldb-10          <  19> send packet: $qRegisterInfo13#a6
lldb-10          < 129> read packet: $name:padding;alt-name:padding;bitsize:32;offset:76;encoding:vector;format:vector-uint8;set:Floating Point Registers;dwarf:19;#74
lldb-10          <  19> send packet: $qRegisterInfo14#a7
lldb-10          < 129> read packet: $name:padding;alt-name:padding;bitsize:32;offset:80;encoding:vector;format:vector-uint8;set:Floating Point Registers;dwarf:20;#67
lldb-10          <  19> send packet: $qRegisterInfo15#a8
lldb-10          < 129> read packet: $name:padding;alt-name:padding;bitsize:32;offset:84;encoding:vector;format:vector-uint8;set:Floating Point Registers;dwarf:21;#6c
lldb-10          <  19> send packet: $qRegisterInfo16#a9
lldb-10          < 129> read packet: $name:padding;alt-name:padding;bitsize:32;offset:88;encoding:vector;format:vector-uint8;set:Floating Point Registers;dwarf:22;#71
lldb-10          <  19> send packet: $qRegisterInfo17#aa
lldb-10          < 129> read packet: $name:padding;alt-name:padding;bitsize:32;offset:92;encoding:vector;format:vector-uint8;set:Floating Point Registers;dwarf:23;#6d
lldb-10          <  19> send packet: $qRegisterInfo18#ab
lldb-10          < 129> read packet: $name:padding;alt-name:padding;bitsize:32;offset:96;encoding:vector;format:vector-uint8;set:Floating Point Registers;dwarf:24;#72
lldb-10          <  19> send packet: $qRegisterInfo19#ac
lldb-10          < 114> read packet: $name:cpsr;alt-name:cpsr;bitsize:32;offset:100;encoding:uint;format:hex;set:General Purpose Registers;dwarf:25;#7f
lldb-10          <  19> send packet: $qRegisterInfo1a#d4
lldb-10          <   7> read packet: $E45#ae
lldb-10          <  16> send packet: $qfThreadInfo#bb
lldb-10          <   7> read packet: $m01#ce
lldb-10          <  16> send packet: $qsThreadInfo#c8
lldb-10          <   5> read packet: $l#6c
lldb-10          <   7> send packet: $Hg1#e0
lldb-10          <   6> read packet: $OK#9a
lldb-10          <   6> send packet: $p0#a0
lldb-10          <   7> read packet: $0*$#7e
lldb-10          <  16> send packet: $qProcessInfo#dc
lldb-10          <   4> read packet: $#00
lldb-10          <  26> send packet: $qStructuredDataPlugins#02
lldb-10          <   4> read packet: $#00
lldb-10          <  13> send packet: $qSymbol::#5b
lldb-10          <   4> read packet: $#00
lldb-10          <  16> send packet: $qfThreadInfo#bb
lldb-10          <   7> read packet: $m01#ce
lldb-10          <  16> send packet: $qsThreadInfo#c8
lldb-10          <   5> read packet: $l#6c
lldb-10          <  16> send packet: $qfThreadInfo#bb
lldb-10          <   7> read packet: $m01#ce
lldb-10          <  16> send packet: $qsThreadInfo#c8
lldb-10          <   5> read packet: $l#6c
lldb-10          <   6> send packet: $pf#d6
lldb-10          <  10> read packet: $0* 5* #f9
lldb-10          <   7> send packet: $p19#da
lldb-10          <  10> read packet: $1000*!#0c
lldb-10          <   6> send packet: $pd#d4
lldb-10          <  10> read packet: $00*!10#0c
lldb-10          <   6> send packet: $pe#d5
lldb-10          <  12> read packet: $78563412#a4
lldb-10          <  30> send packet: $qMemoryRegionInfo:12345678#b8
lldb-10          <   4> read packet: $#00
lldb-10          <   6> send packet: $pb#d2
lldb-10          <   7> read packet: $0*$#7e
dbg.evt-handler  <  16> send packet: $jThreadsInfo#c1
(lldb) dbg.evt-handler  <   4> read packet: $#00
dbg.evt-handler  <  24> send packet: $jThreadExtendedInfo:#b9
dbg.evt-handler  <   4> read packet: $#00
Process 1 stopped
* thread #1, stop reason = signal SIGTRAP
    frame #0: 0x55550000 test.elf`main at test.c:1:12
-> 1   	int main() {
   2   	    int x = 4;
   3   	    int y = 3;
   4
   5   	    x += 1;
   6   	    y += 3;
   7
(lldb) register read
lldb-10          <   6> send packet: $p0#a0
lldb-10          <   7> read packet: $0*$#7e
lldb-10          <   6> send packet: $p1#a1
lldb-10          <   7> read packet: $0*$#7e
lldb-10          <   6> send packet: $p2#a2
lldb-10          <   7> read packet: $0*$#7e
lldb-10          <   6> send packet: $p3#a3
lldb-10          <   7> read packet: $0*$#7e
lldb-10          <   6> send packet: $p4#a4
lldb-10          <   7> read packet: $0*$#7e
lldb-10          <   6> send packet: $p5#a5
lldb-10          <   7> read packet: $0*$#7e
lldb-10          <   6> send packet: $p6#a6
lldb-10          <   7> read packet: $0*$#7e
lldb-10          <   6> send packet: $p7#a7
lldb-10          <   7> read packet: $0*$#7e
lldb-10          <   6> send packet: $p8#a8
lldb-10          <   7> read packet: $0*$#7e
lldb-10          <   6> send packet: $p9#a9
lldb-10          <   7> read packet: $0*$#7e
lldb-10          <   6> send packet: $pa#d1
lldb-10          <   7> read packet: $0*$#7e
lldb-10          <   6> send packet: $pc#d3
lldb-10          <   7> read packet: $0*$#7e
General Purpose Registers:
        r0 = 0x00000000
        r1 = 0x00000000
        r2 = 0x00000000
        r3 = 0x00000000
        r4 = 0x00000000
        r5 = 0x00000000
        r6 = 0x00000000
        r7 = 0x00000000
        r8 = 0x00000000
        r9 = 0x00000000
       r10 = 0x00000000
       r11 = 0x00000000
       r12 = 0x00000000
        sp = 0x10000000
        lr = 0x12345678
        pc = 0x55550000  test.elf`main at test.c:1:12
      cpsr = 0x00000010
```
</details>

### DIFF
--- a/examples/armv4t/gdb/mod.rs
+++ b/examples/armv4t/gdb/mod.rs
@@ -19,6 +19,7 @@ mod extended_mode;
 mod host_io;
 mod memory_map;
 mod monitor_cmd;
+mod register_info_override;
 mod section_offsets;
 mod target_description_xml_override;
 
@@ -112,6 +113,13 @@ impl Target for Emu {
     ) -> Option<
         target::ext::target_description_xml_override::TargetDescriptionXmlOverrideOps<'_, Self>,
     > {
+        Some(self)
+    }
+
+    #[inline(always)]
+    fn support_register_info_override(
+        &mut self,
+    ) -> Option<target::ext::register_info_override::RegisterInfoOverrideOps<'_, Self>> {
         Some(self)
     }
 
@@ -369,6 +377,7 @@ impl target::ext::base::singlethread::SingleThreadRangeStepping for Emu {
 mod custom_arch {
     use core::num::NonZeroUsize;
 
+    use gdbstub::arch::lldb::{Encoding, Format, Generic, Register, RegisterInfo};
     use gdbstub::arch::{Arch, RegId, Registers, SingleStepGdbBehavior};
 
     use gdbstub_arch::arm::reg::id::ArmCoreRegId;
@@ -482,6 +491,98 @@ mod custom_arch {
             Some("never gets returned")
         }
 
+        // (LLDB extension)
+        //
+        // for _purely demonstrative purposes_, even though this provides a working
+        // example, it will get overwritten by RegisterInfoOverride.
+        //
+        // See `examples/armv4t/gdb/register_info_override.rs`
+        fn register_info(reg_id: usize) -> Option<RegisterInfo<'static>> {
+            // Fix for missing 24 => Self::Fps in ArmCoreRegId::from_raw_id
+            let id = if reg_id == 24 { 23 } else { reg_id };
+
+            match ArmCoreRegIdCustom::from_raw_id(id) {
+                Some((_, None)) | None => Some(RegisterInfo::Done),
+                Some((r, Some(size))) => {
+                    let name = match r {
+                        // For the purpose of demonstration, we end the qRegisterInfo packet
+                        // exchange when reaching the Time register id, so that this register can
+                        // only be explicitly queried via the single-register read packet.
+                        ArmCoreRegIdCustom::Time => return Some(RegisterInfo::Done),
+                        ArmCoreRegIdCustom::Core(ArmCoreRegId::Gpr(i)) => match i {
+                            0 => "r0",
+                            1 => "r1",
+                            2 => "r2",
+                            3 => "r3",
+                            4 => "r4",
+                            5 => "r5",
+                            6 => "r6",
+                            7 => "r7",
+                            8 => "r8",
+                            9 => "r9",
+                            10 => "r10",
+                            11 => "r11",
+                            12 => "r12",
+                            _ => "unknown",
+                        },
+                        ArmCoreRegIdCustom::Core(ArmCoreRegId::Sp) => "sp",
+                        ArmCoreRegIdCustom::Core(ArmCoreRegId::Lr) => "lr",
+                        ArmCoreRegIdCustom::Core(ArmCoreRegId::Pc) => "pc",
+                        ArmCoreRegIdCustom::Core(ArmCoreRegId::Fpr(_i)) => "padding",
+                        ArmCoreRegIdCustom::Core(ArmCoreRegId::Fps) => "padding",
+                        ArmCoreRegIdCustom::Core(ArmCoreRegId::Cpsr) => "cpsr",
+                        ArmCoreRegIdCustom::Custom => "custom",
+                        _ => "unknown",
+                    };
+                    let encoding = match r {
+                        ArmCoreRegIdCustom::Core(ArmCoreRegId::Gpr(_i)) => Encoding::Uint,
+                        ArmCoreRegIdCustom::Core(ArmCoreRegId::Sp)
+                        | ArmCoreRegIdCustom::Core(ArmCoreRegId::Pc)
+                        | ArmCoreRegIdCustom::Core(ArmCoreRegId::Cpsr)
+                        | ArmCoreRegIdCustom::Custom => Encoding::Uint,
+                        _ => Encoding::Vector,
+                    };
+                    let format = match r {
+                        ArmCoreRegIdCustom::Core(ArmCoreRegId::Gpr(_i)) => Format::Hex,
+                        ArmCoreRegIdCustom::Core(ArmCoreRegId::Sp)
+                        | ArmCoreRegIdCustom::Core(ArmCoreRegId::Pc)
+                        | ArmCoreRegIdCustom::Core(ArmCoreRegId::Cpsr)
+                        | ArmCoreRegIdCustom::Custom => Format::Hex,
+                        _ => Format::VectorUInt8,
+                    };
+                    let set = match r {
+                        ArmCoreRegIdCustom::Core(ArmCoreRegId::Gpr(_i)) => {
+                            "General Purpose Registers"
+                        }
+                        ArmCoreRegIdCustom::Core(ArmCoreRegId::Sp)
+                        | ArmCoreRegIdCustom::Core(ArmCoreRegId::Pc)
+                        | ArmCoreRegIdCustom::Core(ArmCoreRegId::Cpsr)
+                        | ArmCoreRegIdCustom::Custom => "General Purpose Registers",
+                        _ => "Floating Point Registers",
+                    };
+                    let generic = match r {
+                        ArmCoreRegIdCustom::Core(ArmCoreRegId::Sp) => Some(Generic::Sp),
+                        ArmCoreRegIdCustom::Core(ArmCoreRegId::Pc) => Some(Generic::Pc),
+                        _ => None,
+                    };
+                    let reg = Register {
+                        name,
+                        alt_name: None,
+                        bitsize: (usize::from(size)) * 8,
+                        offset: id * (usize::from(size)),
+                        encoding,
+                        format,
+                        set,
+                        gcc: None,
+                        dwarf: Some(id),
+                        generic,
+                        container_regs: None,
+                        invalidate_regs: None,
+                    };
+                    Some(RegisterInfo::Register(reg))
+                }
+            }
+        }
         // armv4t supports optional single stepping.
         //
         // notably, x86 is an example of an arch that does _not_ support
--- a/src/arch.rs
+++ b/src/arch.rs
@@ -159,6 +159,33 @@ pub trait Arch {
         None
     }
 
+    /// (optional) (LLDB extension) Write register info for one of the arch's
+    /// register.
+    ///
+    /// Implementing this method enables LLDB to dynamically query the
+    /// target's register information one by one.
+    ///
+    /// Some targets don't have register context in the compiled version
+    /// of the debugger. Help the debugger by dynamically supplying the register
+    /// info from the target. The debugger will request the register info in
+    /// a sequential manner till an error packet is received. In LLDB, the
+    /// register info search has the following [order](https://github.com/llvm/llvm-project/blob/369ce54bb302f209239b8ebc77ad824add9df089/lldb/source/Plugins/Process/gdb-remote/ProcessGDBRemote.cpp#L397-L402):
+    ///
+    ///1.    Use the target definition python file if one is specified.
+    ///2.    If the target definition doesn't have any of the info from the
+    ///target.xml (registers) then proceed to read the `target.xml`.
+    ///3.    Fall back on the `qRegisterInfo` packets.
+    ///4.    Use hardcoded defaults if available.
+    ///
+    /// See the LLDB [gdb-remote docs](https://github.com/llvm-mirror/lldb/blob/d01083a850f577b85501a0902b52fd0930de72c7/docs/lldb-gdb-remote.txt#L396)
+    /// for more details on the available information that a single register can
+    /// be described by and [#99](https://github.com/daniel5151/gdbstub/issues/99) for more information on LLDB compatibility.
+    #[inline(always)]
+    fn register_info(reg_id: usize) -> Option<lldb::RegisterInfo<'static>> {
+        let _ = reg_id;
+        None
+    }
+
     /// Encode how the mainline GDB client handles target support for
     /// single-step on this particular architecture.
     ///
@@ -254,4 +281,124 @@ pub enum SingleStepGdbBehavior {
     /// please conduct a test + upstream your findings to `gdbstub_arch`.
     #[doc(hidden)]
     Unknown,
+}
+
+/// LLDB-specific types supporting [`Arch::register_info`] and
+/// [`RegisterInfoOverride`](
+/// crate::target::ext::register_info_override::RegisterInfoOverride) APIs.
+pub mod lldb {
+    /// The architecture's register information of a single register.
+    pub enum RegisterInfo<'a> {
+        /// The register info of a single register that should be written.
+        Register(Register<'a>),
+        /// The `qRegisterInfo` query shall be concluded.
+        Done,
+    }
+
+    /// Describes the register info for a single register of
+    /// the target.
+    pub struct Register<'a> {
+        /// The primary register name.
+        pub name: &'a str,
+        /// An alternate name for the register.
+        pub alt_name: Option<&'a str>,
+        /// Size in bits of a register.
+        pub bitsize: usize,
+        /// The offset within the 'g' and 'G' packet of the register data for
+        /// this register.
+        pub offset: usize,
+        /// The encoding type of the register.
+        pub encoding: Encoding,
+        /// The preferred format for display of this register.
+        pub format: Format,
+        /// The register set name this register belongs to.
+        pub set: &'a str,
+        /// The GCC compiler registers number for this register.
+        /// _Note:_ This denotes the same `KEY:VALUE;` pair as `ehframe:VALUE;`.
+        /// See the LLDB [source](https://github.com/llvm/llvm-project/blob/b92436efcb7813fc481b30f2593a4907568d917a/lldb/source/Plugins/Process/gdb-remote/ProcessGDBRemote.cpp#L493).
+        pub gcc: Option<usize>,
+        /// The DWARF register number for this register that is used for this
+        /// register in the debug information.
+        pub dwarf: Option<usize>,
+        /// Specify as a generic register.
+        pub generic: Option<Generic>,
+        /// Other concrete register values this register is contained in.
+        pub container_regs: Option<&'a [usize]>,
+        /// Specifies which register values should be invalidated when this
+        /// register is modified.
+        pub invalidate_regs: Option<&'a [usize]>,
+    }
+
+    /// Describes the encoding type of the register.
+    #[non_exhaustive]
+    pub enum Encoding {
+        /// Unsigned integer
+        Uint,
+        /// Signed integer
+        Sint,
+        /// IEEE 754 float
+        IEEE754,
+        /// Vector register
+        Vector,
+    }
+
+    /// Describes the preferred format for display of this register.
+    #[non_exhaustive]
+    pub enum Format {
+        /// Binary format
+        Binary,
+        /// Decimal format
+        Decimal,
+        /// Hexadecimal format
+        Hex,
+        /// Floating point format
+        Float,
+        /// 8 bit signed int vector
+        VectorSInt8,
+        /// 8 bit unsigned int vector
+        VectorUInt8,
+        /// 16 bit signed int vector
+        VectorSInt16,
+        /// 16 bit unsigned int vector
+        VectorUInt16,
+        /// 32 bit signed int vector
+        VectorSInt32,
+        /// 32 bit unsigned int vector
+        VectorUInt32,
+        /// 32 bit floating point vector
+        VectorFloat32,
+        /// 128 bit unsigned int vector
+        VectorUInt128,
+    }
+
+    /// Describes the generic types that most CPUs have.
+    #[non_exhaustive]
+    pub enum Generic {
+        /// Program counter register
+        Pc,
+        /// Stack pointer register
+        Sp,
+        /// Frame pointer register
+        Fp,
+        /// Return address register
+        Ra,
+        /// CPU flags register
+        Flags,
+        /// Function argument 1
+        Arg1,
+        /// Function argument 2
+        Arg2,
+        /// Function argument 3
+        Arg3,
+        /// Function argument 4
+        Arg4,
+        /// Function argument 5
+        Arg5,
+        /// Function argument 6
+        Arg6,
+        /// Function argument 7
+        Arg7,
+        /// Function argument 8
+        Arg8,
+    }
 }
--- a/src/protocol/response_writer.rs
+++ b/src/protocol/response_writer.rs
@@ -3,7 +3,8 @@ use alloc::string::String;
 #[cfg(feature = "trace-pkt")]
 use alloc::vec::Vec;
 
-use num_traits::PrimInt;
+use num_traits::identities::one;
+use num_traits::{CheckedRem, PrimInt};
 
 use crate::conn::Connection;
 use crate::internal::BeBytes;
@@ -156,7 +157,7 @@ impl<'a, C: Connection + 'a> ResponseWriter<'a, C> {
     }
 
     /// Write an entire string over the connection.
-    pub fn write_str(&mut self, s: &'static str) -> Result<(), Error<C::Error>> {
+    pub fn write_str(&mut self, s: &str) -> Result<(), Error<C::Error>> {
         for b in s.as_bytes().iter() {
             self.write(*b)?;
         }
@@ -224,6 +225,42 @@ impl<'a, C: Connection + 'a> ResponseWriter<'a, C> {
         let buf = &buf[..len];
         for b in buf.iter().copied().skip_while(|&b| b == 0) {
             self.write_hex(b)?
+        }
+        Ok(())
+    }
+
+    /// Write a number as a decimal string, converting every digit to an ascii
+    /// char.
+    pub fn write_dec<D: PrimInt + CheckedRem>(
+        &mut self,
+        mut digit: D,
+    ) -> Result<(), Error<C::Error>> {
+        if digit.is_zero() {
+            return self.write(b'0');
+        }
+
+        let one: D = one();
+        let ten = (one << 3) + (one << 1);
+        let mut d = digit;
+        let mut pow_10 = one;
+        // Get the number of digits in digit
+        while d >= ten {
+            d = d / ten;
+            pow_10 = pow_10 * ten;
+        }
+
+        // Write every digit from left to right as an ascii char
+        while !pow_10.is_zero() {
+            let mut byte = 0;
+            // We have a single digit here which uses up to 4 bit
+            for i in 0..4 {
+                if !((digit / pow_10) & (one << i)).is_zero() {
+                    byte += 1 << i;
+                }
+            }
+            self.write(b'0' + byte)?;
+            digit = digit % pow_10;
+            pow_10 = pow_10 / ten;
         }
         Ok(())
     }
--- a/src/stub/core_impl/register_info.rs
+++ b/src/stub/core_impl/register_info.rs
@@ -1,0 +1,138 @@
+use super::prelude::*;
+use crate::protocol::commands::ext::RegisterInfo;
+
+use crate::arch::lldb::{Encoding, Format, Generic, Register, RegisterInfo as LLDBRegisterInfo};
+use crate::arch::Arch;
+
+impl<T: Target, C: Connection> GdbStubImpl<T, C> {
+    pub(crate) fn handle_register_info(
+        &mut self,
+        res: &mut ResponseWriter<'_, C>,
+        target: &mut T,
+        command: RegisterInfo,
+    ) -> Result<HandlerStatus, Error<T::Error, C::Error>> {
+        if !target.use_register_info() {
+            return Ok(HandlerStatus::Handled);
+        }
+
+        let handler_status = match command {
+            RegisterInfo::qRegisterInfo(cmd) => {
+                let mut err = Ok(());
+                let cb = &mut |reg: Option<Register<'_>>| {
+                    let res = match reg {
+                        // TODO: replace this with a try block (once stabilized)
+                        Some(reg) => (|| {
+                            res.write_str("name:")?;
+                            res.write_str(reg.name)?;
+                            if let Some(alt_name) = reg.alt_name {
+                                res.write_str(";alt-name:")?;
+                                res.write_str(alt_name)?;
+                            }
+                            res.write_str(";bitsize:")?;
+                            res.write_dec(reg.bitsize)?;
+                            res.write_str(";offset:")?;
+                            res.write_dec(reg.offset)?;
+                            res.write_str(";encoding:")?;
+                            res.write_str(match reg.encoding {
+                                Encoding::Uint => "uint",
+                                Encoding::Sint => "sint",
+                                Encoding::IEEE754 => "ieee754",
+                                Encoding::Vector => "vector",
+                            })?;
+                            res.write_str(";format:")?;
+                            res.write_str(match reg.format {
+                                Format::Binary => "binary",
+                                Format::Decimal => "decimal",
+                                Format::Hex => "hex",
+                                Format::Float => "float",
+                                Format::VectorSInt8 => "vector-sint8",
+                                Format::VectorUInt8 => "vector-uint8",
+                                Format::VectorSInt16 => "vector-sint16",
+                                Format::VectorUInt16 => "vector-uint16",
+                                Format::VectorSInt32 => "vector-sint32",
+                                Format::VectorUInt32 => "vector-uint32",
+                                Format::VectorFloat32 => "vector-float32",
+                                Format::VectorUInt128 => "vector-uint128",
+                            })?;
+                            res.write_str(";set:")?;
+                            res.write_str(reg.set)?;
+                            if let Some(gcc) = reg.gcc {
+                                res.write_str(";gcc:")?;
+                                res.write_dec(gcc)?;
+                            }
+                            if let Some(dwarf) = reg.dwarf {
+                                res.write_str(";dwarf:")?;
+                                res.write_dec(dwarf)?;
+                            }
+                            if let Some(generic) = reg.generic {
+                                res.write_str(";generic:")?;
+                                res.write_str(match generic {
+                                    Generic::Pc => "pc",
+                                    Generic::Sp => "sp",
+                                    Generic::Fp => "fp",
+                                    Generic::Ra => "ra",
+                                    Generic::Flags => "flags",
+                                    Generic::Arg1 => "arg1",
+                                    Generic::Arg2 => "arg2",
+                                    Generic::Arg3 => "arg3",
+                                    Generic::Arg4 => "arg4",
+                                    Generic::Arg5 => "arg5",
+                                    Generic::Arg6 => "arg6",
+                                    Generic::Arg7 => "arg7",
+                                    Generic::Arg8 => "arg8",
+                                })?;
+                            }
+                            if let Some(c_regs) = reg.container_regs {
+                                res.write_str(";container-regs:")?;
+                                res.write_num(c_regs[0])?;
+                                for reg in c_regs.iter().skip(1) {
+                                    res.write_str(",")?;
+                                    res.write_num(*reg)?;
+                                }
+                            }
+                            if let Some(i_regs) = reg.invalidate_regs {
+                                res.write_str(";invalidate-regs:")?;
+                                res.write_num(i_regs[0])?;
+                                for reg in i_regs.iter().skip(1) {
+                                    res.write_str(",")?;
+                                    res.write_num(*reg)?;
+                                }
+                            }
+                            res.write_str(";")
+                        })(),
+                        // In fact, this doesn't has to be E45! It could equally well be any
+                        // other error code or even an eOk, eAck or eNack! It turns out that
+                        // 0x45 == 69, so presumably the LLDB people were just having some fun
+                        // here. For a little discussion on this and LLDB source code pointers,
+                        // see https://github.com/daniel5151/gdbstub/pull/103#discussion_r888590197
+                        _ => res.write_str("E45"),
+                    };
+                    if let Err(e) = res {
+                        err = Err(e);
+                    }
+                };
+                if let Some(ops) = target.support_register_info_override() {
+                    use crate::target::ext::register_info_override::{Callback, CallbackToken};
+
+                    ops.register_info(
+                        cmd.reg_id,
+                        Callback {
+                            cb,
+                            token: CallbackToken(core::marker::PhantomData),
+                        },
+                    )
+                    .map_err(Error::TargetError)?;
+                    err?;
+                } else if let Some(reg) = T::Arch::register_info(cmd.reg_id) {
+                    match reg {
+                        LLDBRegisterInfo::Register(reg) => cb(Some(reg)),
+                        LLDBRegisterInfo::Done => cb(None),
+                    };
+                }
+                HandlerStatus::Handled
+            }
+        };
+
+        Ok(handler_status)
+    }
+}
--- a/src/target/ext/register_info_override.rs
+++ b/src/target/ext/register_info_override.rs
@@ -1,0 +1,54 @@
+//! (LLDB extension) Override the register info specified by `Target::Arch`.
+use crate::arch::lldb::Register;
+use crate::target::Target;
+
+/// If your target is using the `qRegisterInfo` packet, this
+/// token hast to be returned from the `register_info` function to guarantee
+/// that the callback function to write the register info has been invoked.
+pub struct CallbackToken<'a>(pub(crate) core::marker::PhantomData<&'a *mut ()>);
+
+/// This struct is used internally by `gdbstub` to wrap a
+/// callback function which has to be used to direct the `qRegisterInfo` query.
+pub struct Callback<'a> {
+    /// The callback function that is directing the `qRegisterInfo` query.
+    pub(crate) cb: &'a mut dyn FnMut(Option<Register<'_>>),
+    /// A token to guarantee the callback has been used.
+    pub(crate) token: CallbackToken<'a>,
+}
+
+impl<'a> Callback<'a> {
+    /// The `qRegisterInfo` query shall be concluded.
+    #[inline(always)]
+    pub fn done(self) -> CallbackToken<'a> {
+        (self.cb)(None);
+        self.token
+    }
+
+    /// Write the register info of a single register.
+    #[inline(always)]
+    pub fn write(self, reg: Register<'_>) -> CallbackToken<'a> {
+        (self.cb)(Some(reg));
+        self.token
+    }
+}
+
+/// Target Extension - Override the target register info
+/// specified by `Target::Arch`.
+///
+/// _Note:_ Unless you're working with a particularly dynamic,
+/// runtime-configurable target, it's unlikely that you'll need to implement
+/// this extension.
+pub trait RegisterInfoOverride: Target {
+    /// Invoke `reg_info.write(reg)` where `reg` is a [`Register`
+    /// ](crate::arch::lldb::Register) struct to write information of
+    /// a single register or `reg_info.done()` if you want to end the
+    /// `qRegisterInfo` packet exchange. These two methods will return a
+    /// `CallbackToken`, which has to be returned from this method.
+    fn register_info<'a>(
+        &mut self,
+        reg_id: usize,
+        reg_info: Callback<'a>,
+    ) -> Result<CallbackToken<'a>, Self::Error>;
+}
+
+define_ext!(RegisterInfoOverrideOps, RegisterInfoOverride);
--- a/src/target/mod.rs
+++ b/src/target/mod.rs
@@ -597,6 +597,21 @@ pub trait Target {
         true
     }
 
+    /// (LLDB extension) Whether to send register information to the client.
+    ///
+    /// Setting this to `false` will override both
+    /// [`Target::support_register_info_override`] and the associated
+    /// [`Arch::register_info`].
+    ///
+    /// _Author's note:_ Having the LLDB client autodetect your target's
+    /// register set is really useful, so unless you're _really_ trying to
+    /// squeeze `gdbstub` onto a particularly resource-constrained platform, you
+    /// may as well leave this enabled.
+    #[inline(always)]
+    fn use_register_info(&self) -> bool {
+        true
+    }
+
     /// Support for setting / removing breakpoints.
     #[inline(always)]
     fn support_breakpoints(&mut self) -> Option<ext::breakpoints::BreakpointsOps<'_, Self>> {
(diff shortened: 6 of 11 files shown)


daniel5151:
Hey there, thanks for sending this in!

I'm actually on vacation right now, so I won't get the chance to give this a thorough review until sometime next week. In the meantime, I'll check in occasionally to approve CI runs, since the first and most obvious thing you'll need to fix is that the code within `gdbstub` must be no_std (i.e: no using `Vec`). 

I would suggest using the `./example_no_std/check_size.sh` script to ensure the code is no_std compatible, and also make sure that the implementation doesn't introduce exorbitant binary size overhead on non-lldb implementations.

jawilk:
Thanks for the quick reply and apologies for disturbing your vacation with this :D

Looks like I lost the view of the bigger picture somehow, it should be no_std now.
The `write_dec()` method is weighting in with
```
File  .text    Size             Crate Name
0.0%   0.7%    131B           gdbstub gdbstub::protocol::response_writer::ResponseWriter<C>::write_dec
```
compared to the existing ones
```
0.1%   1.2%    207B           gdbstub gdbstub::protocol::response_writer::ResponseWriter<C>::write
0.0%   0.7%    130B           gdbstub gdbstub::protocol::response_writer::ResponseWriter<C>::inner_write
0.0%   0.6%    115B           gdbstub gdbstub::protocol::response_writer::ResponseWriter<C>::flush
0.0%   0.5%     88B           gdbstub gdbstub::protocol::response_writer::ResponseWriter<C>::write_hex
0.0%   0.6%    104B           gdbstub gdbstub::protocol::response_writer::ResponseWriter<C>::write_num
```

<details>
<summary>cargo bloat output</summary>

```
File  .text    Size             Crate Name
3.7%  72.2% 12.7KiB         [Unknown] main
0.2%   4.4%    797B           gdbstub gdbstub::stub::state_machine::GdbStubStateMachineInner<gdbstub::stub::state_machine::state::Running,T,C>::report_stop
0.1%   2.0%    366B           gdbstub gdbstub::protocol::commands::breakpoint::BasicBreakpoint::from_slice
0.1%   1.6%    295B    gdbstub_nostd? <gdbstub_nostd::gdb::DummyTarget as gdbstub::target::ext::register_info::RegisterInfo>::get_register_info
0.1%   1.5%    268B           gdbstub gdbstub::protocol::common::hex::decode_hex_buf
0.1%   1.4%    253B           gdbstub gdbstub::protocol::response_writer::ResponseWriter<C>::write_specific_thread_id
0.1%   1.4%    249B           gdbstub gdbstub::stub::core_impl::resume::<impl gdbstub::stub::core_impl::GdbStubImpl<T,C>>::write_stop_common
0.1%   1.3%    232B           gdbstub <gdbstub::protocol::common::thread_id::ThreadId as core::convert::TryFrom<&[u8]>>::try_from
0.1%   1.2%    207B           gdbstub gdbstub::protocol::response_writer::ResponseWriter<C>::write
0.0%   0.7%    133B           gdbstub gdbstub::protocol::common::hex::decode_hex
0.0%   0.7%    133B           gdbstub gdbstub::protocol::common::hex::decode_hex
0.0%   0.7%    131B           gdbstub gdbstub::protocol::response_writer::ResponseWriter<C>::write_dec
0.0%   0.7%    130B           gdbstub gdbstub::protocol::response_writer::ResponseWriter<C>::inner_write
0.0%   0.6%    115B           gdbstub gdbstub::protocol::response_writer::ResponseWriter<C>::flush
0.0%   0.6%    114B           gdbstub gdbstub::protocol::common::hex::decode_hex
0.0%   0.6%    108B           gdbstub gdbstub::protocol::response_writer::ResponseWriter<C>::write_num
0.0%   0.6%    104B           gdbstub gdbstub::protocol::response_writer::ResponseWriter<C>::write_num
0.0%   0.6%    103B           gdbstub gdbstub::protocol::response_writer::ResponseWriter<C>::write_num
0.0%   0.6%    102B    gdbstub_nostd? <gdbstub_nostd::gdb::DummyTarget as gdbstub::target::ext::base::multithread::MultiThreadBase>::read_addrs
0.0%   0.6%    101B         [Unknown] __libc_csu_init
0.0%   0.5%     92B           gdbstub <usize as gdbstub::internal::be_bytes::BeBytes>::from_be_bytes
0.0%   0.5%     90B           gdbstub <u32 as gdbstub::internal::be_bytes::BeBytes>::from_be_bytes
0.0%   0.5%     88B           gdbstub gdbstub::protocol::response_writer::ResponseWriter<C>::write_hex
0.0%   0.4%     70B           gdbstub <gdbstub::protocol::common::thread_id::IdKind as core::convert::TryFrom<&[u8]>>::try_from
0.0%   0.4%     65B      gdbstub_arch <gdbstub_arch::arm::reg::arm_core::ArmCoreRegs as gdbstub::arch::Registers>::gdb_deserialize
0.0%   0.4%     65B    gdbstub_nostd? <gdbstub_nostd::gdb::DummyTarget as gdbstub::target::ext::base::multithread::MultiThreadBase>::write_addrs
0.0%   0.4%     65B    gdbstub_nostd? <gdbstub_nostd::gdb::DummyTarget as gdbstub::target::ext::base::multithread::MultiThreadBase>::write_registers
0.0%   0.4%     65B    gdbstub_nostd? <gdbstub_nostd::gdb::DummyTarget as gdbstub::target::ext::base::multithread::MultiThreadBase>::read_registers
0.0%   0.3%     57B compiler_builtins __rust_probestack
0.0%   0.3%     50B    gdbstub_nostd? <gdbstub_nostd::gdb::DummyTarget as gdbstub::target::ext::base::multithread::MultiThreadResume>::set_resume_action_continue
0.0%   0.3%     50B    gdbstub_nostd? <gdbstub_nostd::gdb::DummyTarget as gdbstub::target::ext::base::multithread::MultiThreadResume>::clear_resume_actions
0.0%   0.3%     50B    gdbstub_nostd? <gdbstub_nostd::gdb::DummyTarget as gdbstub::target::ext::base::multithread::MultiThreadResume>::resume
0.0%   0.2%     43B         [Unknown] _start
0.0%   0.0%      6B    gdbstub_nostd? <gdbstub_nostd::gdb::DummyTarget as gdbstub::target::ext::breakpoints::SwBreakpoint>::remove_sw_breakpoint
0.0%   0.0%      6B    gdbstub_nostd? <gdbstub_nostd::gdb::DummyTarget as gdbstub::target::ext::breakpoints::SwBreakpoint>::add_sw_breakpoint
0.0%   0.0%      2B         [Unknown] __libc_csu_fini
0.0%   0.0%      0B                   And 0 smaller methods. Use -n N to show more.
5.2% 100.0% 17.6KiB                   .text section size, the file size is 339.5KiB
section               size      addr
.interp                 28       568
.note.ABI-tag           32       596
.note.gnu.build-id      36       628
.gnu.hash               28       664
.dynsym                360       696
.dynstr                193      1056
.gnu.version            30      1250
.gnu.version_r          48      1280
.rela.dyn             1032      1328
.init                   23      2360
.plt                    16      2384
.plt.got                 8      2400
.text                17986      2416
.fini                    9     20404
.rodata               1130     20416
.eh_frame_hdr          316     21548
.eh_frame             1616     21864
.init_array              8   2120712
.fini_array              8   2120720
.data.rel.ro           416   2120728
.dynamic               448   2121144
.got                   136   2121592
.data                    8   2121728
.bss                     8   2121736
.comment                41         0
Total                23964
```
</details>

jawilk:
Thanks a lot again for taking the time, your explanations and pointers were very helpful!

- `write_dec` is using generics now
- I switched to a callback function in `get_register_info`, but the only way I could make it work was to wrap the callback into a Box `Box<dyn FnOnce(Option<Register<'_>>) + '_>` which fails for non_std. I need to think a bit more about that.

Regarding changes to the current code, I changed the lifetime of the `s` argument in `write_str` to use an anonymous lifetime instead of a static one
```
pub fn write_str(&mut self, s: &'_ str) -> Result<(), Error<C::Error>> {
```

daniel5151:
Ah, yes, as for FnOnce, things get a bit complicated, since it needs to take `self` when invoked, and you can't really do that (in safe Rust) with a stack allocated closure...

Fortunately, this situation is similar to one we encountered while working on Host IO support, and though we didn't end up using a callback for that API, I have a feeling using a callback in this API is a good idea.

Please check out this comment for a proposed API: https://github.com/daniel5151/gdbstub/pull/66#discussion_r683603624
The linked Rust Playground gist is pretty much what you'd want to implement here. i.e: instead of using a bare `&mut dyn FnMut(..)`, you'd wrap that callback in a struct that provides a method that forwards to that implementation, while also consuming self in the process + returning a ZST "token" that must be returned from the handler, thereby guaranteeing that the function has been called.

If you can pull that off, that should get the CI green.

src/protocol/response_writer.rs
```diff
@@ -156,7 +157,7 @@ impl<'a, C: Connection + 'a> ResponseWriter<'a, C> {
     }
 
     /// Write an entire string over the connection.
-    pub fn write_str(&mut self, s: &'static str) -> Result<(), Error<C::Error>> {
+    pub fn write_str(&mut self, s: &'_ str) -> Result<(), Error<C::Error>> {
```
daniel5151, line 160: this can just be `&str`

src/target/ext/register_info.rs
```diff
@@ -0,0 +1,49 @@
...
+use crate::target::Target;
+
+/// Describes the register info for a single register of the target
+pub struct Register<'a> {
+    /// The primary register name
+    pub name: &'a str,
+    /// An alternate name for the register
+    pub alt_name: &'a str,
+    /// Size in bits of a register (base 10)
+    pub bitsize: u16,
```
daniel5151, line 16: as mentioned earlier, unless the spec specifies a specific bitsize, please default to using `usize`. This applies to most `u16` fields in this struct.

src/target/ext/register_info.rs
```diff
@@ -0,0 +1,49 @@
...
+    /// The offset within the 'g' and 'G' packet of the register data for this register (base 10)
+    pub offset: u16,
+    /// The encoding type of the register
+    pub encoding: &'a str,
+    /// The preferred format for display of this register
+    pub format: &'a str,
```
daniel5151, line 22: reading the spec, it seems that `encoding` and `format` should both use `#[non_exhaustive] enum`, rather than being raw `str`s.

src/target/ext/register_info.rs
```diff
@@ -0,0 +1,49 @@
...
+    /// The register set name this register belongs to
+    pub set: &'a str,
+    /// The GCC compiler registers number for this register
+    pub gcc: Option<u16>,
+    /// The DWARF register number for this register that is used for this register in the debug information
+    pub dwarf: Option<u16>,
+    /// Specify as a generic register
+    pub generic: Option<&'a str>,
```
daniel5151, line 30: this should also be an `#[non_exhaustive] enum`

src/target/ext/register_info.rs
```diff
@@ -0,0 +1,49 @@
...
+    /// Other concrete register values this register is contained in (expecting base 10 here but will be written with write_num() as hex)
```
daniel5151, line 31: no need to include protocol nitty gritty details in the help text here (i.e: the end user doesn't care if it's being written in decimal vs. hex)

src/target/ext/register_info.rs
```diff
@@ -0,0 +1,49 @@
...
+    pub container_regs: Option<&'a [u16]>,
+    /// Specifies which register values should be invalidated when this register is modified (expecting base 10 here but will be written with write_num() as hex)
+    pub invalidate_regs: Option<&'a [u16]>,
+}
+
+/// Target Extension - Get register information from the target.
+///
+/// Corresponds to the `qRegisterInfo` command.
+pub trait RegisterInfo: Target {
+    /// Return the target's register info for a single register or Ok(None) if <reg_id> is greater than the number of registers to end the request.
```
daniel5151, line 41: if you take my aformentioned advice on introducing a wrapper struct around the callback, it _might_ be possible to have a separate `.done()` method, rather than using an `Option` here. I say _might_, since the lifetimes might make things a bit complicated

src/target/mod.rs
```diff
@@ -665,6 +665,13 @@ pub trait Target {
     fn support_auxv(&mut self) -> Option<ext::auxv::AuxvOps<'_, Self>> {
         None
     }
+
+    /// Support for querying register info in LLDB.
+    /// Let the debugger request register info one by one. Should be used if 'use_target_description_xml()' is disabled.
```
daniel5151, line 670: nit: add a newline between the brief text and the expanded text, or else rustdoc will put everything on one line.

daniel5151 (changes requested):
Found some time tonight to give the code a proper comb-over. Left a bunch of comments. Overall, things are looking good!

src/target/ext/register_info.rs
```diff
@@ -0,0 +1,49 @@
...
+//! _Note:_ LLDB specific! This packet is meant as an alternative to the 'qXfer:features:read:target.xml' packet. LLDB will fall back to the .xml file if qRegisterInfo isn't supported on the target. See issue [#99](https://github.com/daniel5151/gdbstub/issues/99) for more info on LLDB compatibility.
+//!
+use crate::target::Target;
+
+/// Describes the register info for a single register of the target
+pub struct Register<'a> {
+    /// The primary register name
+    pub name: &'a str,
+    /// An alternate name for the register
+    pub alt_name: &'a str,
```
daniel5151, line 14: seems like this field is optional, at least according to the linked lldb spec doc

src/stub/core_impl/register_info.rs
```diff
@@ -0,0 +1,105 @@
...
+                                        write_register!(c_regs, ";container-regs:");
+                                    }
+                                    if let Some(i_regs) = invalidate_regs {
+                                        write_register!(i_regs, ";invalidate-regs:");
+                                    }
+                                    res.write_str(";")
+                                })()
+                            }
+                            // This will end the 'qRegisterInfo' query
+                            None => res.write_str("E45"),
```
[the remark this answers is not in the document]
jawilk: The `qRegisterInfo` request loop starts [here](https://github.com/llvm/llvm-project/blob/369ce54bb302f209239b8ebc77ad824add9df089/lldb/source/Plugins/Process/gdb-remote/ProcessGDBRemote.cpp#L444) and goes on till
`response_type != StringExtractorGDBRemote::eResponse`. It will get reassigned in the loop by `response.GetResponseType();` which is defined [here](https://github.com/llvm/llvm-project/blob/369ce54bb302f209239b8ebc77ad824add9df089/lldb/source/Utility/StringExtractorGDBRemote.cpp#L18) and returns `eError` if we parse an `E` and the packet has size 3
```
case 'E':
    if (isxdigit(m_packet[1]) && isxdigit(m_packet[2])) {
      if (m_packet.size() == 3)
        return eError;
```
So we could indeed return something else here (maybe a unique value to gdbstub?) :laughing:

src/protocol/response_writer.rs
```diff
@@ -228,6 +229,34 @@ impl<'a, C: Connection + 'a> ResponseWriter<'a, C> {
...
+        let mut d = digit;
+        let mut pow_10 = _1;
+        // Get the number of digits in digit
+        while d >= _10 {
+            d = d / _10;
+            pow_10 = pow_10 * _10;
+        }
+        // Write every digit from left to right as an ascii char
+        while !pow_10.is_zero() {
+            self.write(b'0' + (digit / pow_10).to_u8().unwrap())?;
```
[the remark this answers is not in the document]
jawilk: since the generic is down to a single digit here I changed this to extract the 4 LSBs directly, the unwrap is gone now but `write_dec` got a bit lenghty

src/target/ext/register_info.rs
jawilk: oh yes a `.done()` method would make it way cleaner. There is one now but I only moved the `Option` into the callback argument itself since I had some trouble accessing the `ResponseWriter` from anywhere else :thinking: 
```
pub struct Callback<'a> {
    /// The callback function that has to be called exactly once
    pub cb: &'a mut dyn FnMut(Option<Register<'_>>),
    /// The "proof token" for the callback invocation
    pub token: CallbackToken<'a>,
}
```

src/target/ext/register_info.rs
```diff
@@ -0,0 +1,220 @@
...
+    Sint,
+    /// IEEE 754 float
+    IEEE754,
+    /// vector register
+    Vector,
+}
+
+impl Encoding {
+    /// Match an enum field to its corresponding string slice
+    pub fn as_str(&self) -> &str {
```
daniel5151, line 65: Unless I'm missing something, I don't see why an end user would ever want to obtain the specific backing string corresponding to an enum variant. I suspect these functions should just be inline `match` statements within the core_impl code, and not part of the public API

src/target/ext/register_info.rs
```diff
@@ -0,0 +1,220 @@
...
+            Generic::Arg6 => "arg6",
+            Generic::Arg7 => "arg7",
+            Generic::Arg8 => "arg8",
+        }
+    }
+}
+
+/// This Zero-Sized Type (ZST) token has to be returned from the callback in the
+/// Callback struct below. This is imitating the FnOnce trait.
+pub struct CallbackToken<'a>(pub core::marker::PhantomData<&'a *mut ()>);
```
daniel5151, line 179: Inner field should be `pub(crate)`, not `pub`

src/target/ext/register_info.rs
```diff
@@ -0,0 +1,220 @@
...
+
+/// Wrapper for a callback. The <token> field is our "proof-of-invocation", to
+/// guarantee that the callback stored in the <cb> member has been called.
+pub struct Callback<'a> {
+    /// The callback function that has to be called exactly once
+    pub cb: &'a mut dyn FnMut(Option<Register<'_>>),
```
daniel5151, line 185: ditto on using `pub(crate)` for the inner fields

src/target/ext/register_info.rs
```diff
@@ -0,0 +1,220 @@
...
+    /// The "proof token" for the callback invocation
+    pub token: CallbackToken<'a>,
+}
+
+/// Trait which acts as a forwarder to the callback in the <cb> field. It will
```
daniel5151, line 190: Let's workshop this (and other) doc comments to avoid referencing underlying implementation details, and focusing on how/why they are used by the end user.

src/target/ext/register_info.rs
```diff
@@ -0,0 +1,220 @@
...
+/// consume 'self' and return the token to guarantee the callback has been
+/// called.
+impl<'a> Callback<'a> {
+    /// The 'qRegisterInfo' query shall be concluded
+    pub fn done(self) -> CallbackToken<'a> {
```
daniel5151, line 195: nit: likely unnecessary, but throw some #[inline(always)] on these functions

src/target/ext/register_info.rs
```diff
@@ -0,0 +1,220 @@
...
+        (self.cb)(Some(reg));
+        self.token
+    }
+}
+
+/// Target Extension - Get register information from the target.
+///
+/// Corresponds to the `qRegisterInfo` command.
+pub trait RegisterInfo: Target {
+    /// Return the target's register info for a single register or Ok(None) if
```
daniel5151, line 211: Stale comment

daniel5151 (changes requested):
Some more comments...

Also, on a broader note: I think this new way of reporting registers should also become an (optional) part of the `Arch` trait, akin to how target.xml is handled today. Of course, you'll have to implement it in such a way that if neither the Arch not the Target provide an implementation, the code still gets dead code eliminated...

If you can take a crack at that, that'd be much appreciated. If not, I can take a hand at helping out with that sometime next week.

src/target/ext/register_info.rs
```diff
@@ -0,0 +1,220 @@
...
+    /// Binary format
+    Binary,
+    /// Decimal format
+    Decimal,
+    /// Hexadecimal format
+    Hex,
+    /// Floating point format
+    Float,
+    /// 8 bit signed int vector
+    VectorOfSInt8,
```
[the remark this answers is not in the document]
jawilk: was taking the names from the lldb [source](https://github.com/llvm/llvm-project/blob/369ce54bb302f209239b8ebc77ad824add9df089/lldb/tools/debugserver/source/DNBDefs.h#L177), will remove the "Of"s I also think it's not needed/too verbose

jawilk:
> If you can take a crack at that, that'd be much appreciated. If not, I can take a hand at helping out with that sometime next week.

I'll give it a try tomorrow! I was actually planning to try upstream the Berkeley Packet Filter (BPF) [arch](https://github.com/jawilk/gdbstub/tree/bpf/gdbstub_arch/src/bpf) so getting a better sense on that side of things is a good idea

src/target/ext/register_info.rs
daniel5151: Sure, that seems reasonable. The key thing is that the end user API doesn't have to do that explicit `None` `Some` wrapping, making the intent more clear :)

src/protocol/response_writer.rs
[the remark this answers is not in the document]
daniel5151: I might end up taking a crack at writing an implementation myself at some point, but at first blush, what you've got now seems fine. I know how annoying it can be doing generic math in Rust, so having this be a wordy function might just be how it is ¯\\\_(ツ)_/¯

src/stub/core_impl/register_info.rs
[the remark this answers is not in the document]
daniel5151: Haha, nice. Honestly, I'd just add a little inline comment summarising what we've discovered here, and leave the value as is. After all, I think it's kinda funny lol

src/target/mod.rs
```diff
@@ -665,6 +665,15 @@ pub trait Target {
     fn support_auxv(&mut self) -> Option<ext::auxv::AuxvOps<'_, Self>> {
         None
     }
+
+    /// Support for querying register info in LLDB.
```
daniel5151, line 669:
```suggestion
    /// (LLDB specific) Alternate method of querying register info.
```

I haven't decided on how `gdbstub`'s long-term LLDB compat story is going to look like, but for now, I'd lean towards marking any "entrypoints" into LLDB-specific sections of the codebase with `(LLDB specific)`. This applies to the mod-comment in `register_info.rs` as well.

daniel5151:
> > If you can take a crack at that, that'd be much appreciated. If not, I can take a hand at helping out with that sometime next week.
> 
> I'll give it a try tomorrow! I was actually planning to try upstream the Berkeley Packet Filter (BPF) [arch](https://github.com/jawilk/gdbstub/tree/bpf/gdbstub_arch/src/bpf) so getting a better sense on that side of things is a good idea

Oh, awesome! Always happy to see more `Arch` implementations get upstreamed.
Of course, as a reminder, we'd want to land that as a separate PR, in case you were thinking of lumping it in with this one :)

As for the implementation: you'll probably want to mirror how `target.xml` is being handled:

- Add a method akin to `Target::use_target_description_xml`, which can be used to globally enable/disable the use of this functionality
- Add a new method to `Arch`, except instead of returning a bare `CallbackToken`, have it return a `Option<CallbackToken>`, where the default implementation returns `None`
- Within `gdbstub`, keep a close eye on ensuring that dead-code-elimination still kicks in when neither the `Arch` nor the `Target` provide an implementation of this handler (or the aforementioned global enable/disable `use_` method is set to `false`)

If you can't seem to get things working, just let me know. `gdbstub` has some pretty unique constraints compared to typical Rust projects, so I'm more than happy to do some hands on work to get this PR landed!

src/target/mod.rs
daniel5151: Ahh, sorry. It seems that the wording that's currently being used is actually `(LLDB extension)`, so please use that instead.

See https://github.com/daniel5151/gdbstub/pull/100#discussion_r811520963 for the last time this came up.

jawilk:
Seems like `24 => Self::Fps,` is missing here

https://github.com/daniel5151/gdbstub/blob/1fde6f720f5928b2b47229e8ff9e88ed6fea82b8/gdbstub_arch/src/arm/reg/id.rs#L27-L35

examples/armv4t/gdb/mod.rs
```diff
@@ -80,6 +80,16 @@ impl Target for Emu {
...
+    #[inline(always)]
+    fn use_target_description_xml(&self) -> bool {
+        true
+    }
+
+    #[inline(always)]
+    fn use_register_info(&self) -> bool {
+        true
+    }
+
```
daniel5151, line 92: you can omit these lines - the functions default to `true`, so there's no need to explicitly redefine them here.

examples/armv4t/gdb/mod.rs
```diff
@@ -482,6 +495,76 @@ mod custom_arch {
             Some("never gets returned")
         }
 
+        // (LLDB extension) TODO
+        fn register_info<'a>(reg_id: usize, write_info: Callback<'a>) -> Option<CallbackToken<'a>> {
+            match ArmCoreRegId::from_raw_id(reg_id) {
```
daniel5151, line 500: this should be using `ArmCoreRegIdCustom`, and include `Custom` register (but not the `Time` register). See the implementation of target.xml override for Armv4tCustom for details

src/arch.rs
```diff
@@ -255,3 +286,144 @@ pub enum SingleStepGdbBehavior {
     #[doc(hidden)]
     Unknown,
 }
+
+/// (LLDB extension) Describes the register info for a single register of the
```
daniel5151, line 290: please hide these new types behind a `pub mod lldb` module to avoid cluttering the `arch` module with a bunch of new lldb-specific types.

src/arch.rs
```diff
@@ -159,6 +159,37 @@ pub trait Arch {
         None
     }
 
+    /// (optional) (LLDB extension) Return a `CallbackToken` that confirms that
+    /// the arch's register info of a single register has been send.
+    ///
+    /// Implementing this method enables LLDB to dynamically query the
+    /// target's register information one by one.
+
+    /// Some targets don't have register
```
daniel5151, line 168: nit: a little bit of word wrap formatting issues here

src/arch.rs
```diff
@@ -159,6 +159,37 @@ pub trait Arch {
         None
     }
 
+    /// (optional) (LLDB extension) Return a `CallbackToken` that confirms that
+    /// the arch's register info of a single register has been send.
+    ///
+    /// Implementing this method enables LLDB to dynamically query the
+    /// target's register information one by one.
+
```
daniel5151, line 167: nit: for consistency with the rest of the codebase, include empty `///` lines between blocks of comments

src/arch.rs
```diff
@@ -159,6 +159,37 @@ pub trait Arch {
...
+    /// for more details on the available information that a single register can
+    /// be described by and [#99](https://github.com/daniel5151/gdbstub/issues/99) for more information on LLDB compatibility.
+
+    /// Invoke `write_info.write(reg)` where `reg` is the `Register` struct from
+    /// this module to write information of a single register or
+    /// `write_info.done()` if you want to end the `qRegisterInfo` packet
+    /// exchange. These two methods will return a `CallbackToken`, which has to
+    /// be returned from this function.
+    #[inline(always)]
+    fn register_info(_reg_id: usize, _write_info: Callback<'_>) -> Option<CallbackToken<'_>> {
```
daniel5151, line 189: nit: avoid using `_name` in trait defns, as this leaks into the rustdoc for the signature. Instead, just add a `let _ = (reg_id, write_info);` within the function body to squelch the warning.

daniel5151, line 189: I would actually change this function signature to `fn register_info(reg_id: usize) -> Option<RegisterInfo<'static>>`. After all, the `Arch` trait is essentially a compile-time container for all the statically known bits of info about an `Arch`, which includes register info, so using a `'static` lifetime is completely appropriate.

If a user wishes to dynamically configure / override this data, they should use the corresponding `Target::supports_lldb_register_info_override` IDET.

* * *

Note that in order to handle the tri-state of "register info, done reporting, unimplemented", I am suggesting that you introduce a new type: 

```rust
enum RegisterInfo<'a> {
    Register(Register<'a>),
    Done
}
```

Which then gets wrapped in an `Option` to denote the "not-implemented" case.

This is the "static" equivalent of the "dynamic" version of this functionality, where an IDET handles the "not-implemented" case, and the two callback methods correspond to returning a `Register` or reporting `Done`.

src/arch.rs
```diff
@@ -255,3 +286,144 @@ pub enum SingleStepGdbBehavior {
...
+    /// Function argument 7
+    Arg7,
+    /// Function argument 8
+    Arg8,
+}
+
+/// (LLDB extension) If your target is using the `qRegisterInfo` packet, this
+/// token hast to be returned from the `register_info` function to guarantee
+/// that the callback function to send the register info has been invoked.
+pub struct CallbackToken<'a>(pub(crate) core::marker::PhantomData<&'a *mut ()>);
```
daniel5151, line 399: `Callback` and `CallbackToken` can move back into the `target` module, as they aren't used by the `Arch`-level implementation of this functionality.

src/target/mod.rs
```diff
@@ -597,6 +597,12 @@ pub trait Target {
         true
     }
 
+    /// Support for register info query
```
daniel5151, line 600: nit: missing `(LLDB extension)`

src/target/mod.rs
```diff
@@ -597,6 +597,12 @@ pub trait Target {
...
+    #[inline(always)]
+    fn use_register_info(&self) -> bool {
+        false
```
daniel5151, line 603: this should default to `true`.

src/target/mod.rs
daniel5151, line 600: i'd just copy over the comment from `use_target_description_xml` wholesale, swapping out the methods with those associated with `qRegisterInfo`

daniel5151 (changes requested):
So, when I said that you should implement this at the `Arch` level, I'd meant that _in addition_ to the dynamic `Target`-level implementation 😅 

Please re-add the dynamic override implementation, as it is useful for implementations that are generic across multiple architectures.

i.e: there should be 3 "knobs" for specifying qRegisterInfo data:

- the top-level `use_register_info` override, that enables completely disabling the packet (regardless of whether the `Arch` or `Target` includes implementations of it)
- The `Arch` level _static_ Register information
- The `Target` level _dynamic_ "override" Register information

Like I said - look at how target.xml is being handled, and match that logic, fallbacks + overrides and all.

daniel5151:
> Seems like `24 => Self::Fps,` is missing here

huh, would you look at that 👀 

I'll try and remember to push up a hotfix right to `master` at some point

src/stub/core_impl/register_info.rs
```diff
@@ -0,0 +1,134 @@
...
+                                res.write_str(";invalidate-regs:")?;
+                                res.write_num(i_regs[0])?;
+                                for reg in i_regs.iter().skip(1) {
+                                    res.write_str(",")?;
+                                    res.write_num(*reg)?;
+                                }
+                            }
+                            res.write_str(";")
+                        })(),
+                        // In fact, this doesn't has to be E45! It could equally well be any other error code or even an eOk, eAck or eNack! It turns out that 0x45 == 69, so presumably the LLDB people were just having some fun here. For a little discussion on this and LLDB source code pointers, see https://github.com/daniel5151/gdbstub/pull/103#discussion_r888590197
```
daniel5151, line 103: nit: run nightly rustfmt manually. sorry about that...

src/target/ext/register_info_override.rs
```diff
@@ -0,0 +1,59 @@
...
+
+/// (LLDB extension) Target Extension - Override the target register info
+/// specified by `Target::Arch`.
+///
+/// _Note:_ Unless you're working with a particularly dynamic,
+/// runtime-configurable target, it's unlikely that you'll need to implement
+/// this extension.
+pub trait RegisterInfoOverride: Target {
+    /// Invoke `reg_info.write(reg)` where `reg` is a [Register
+    /// struct](../../../arch/lldb/struct.Register.html) to write information of
```
daniel5151, line 48: nit: please use intradoc links rahter than linking to HTML directly: https://rust-lang.github.io/rfcs/1946-intra-rustdoc-links.html

src/target/ext/register_info_override.rs
```diff
@@ -0,0 +1,59 @@
+//! (LLDB extension) Override the register info specified by `Target::Arch`.
+use crate::arch::lldb::Register;
+use crate::target::Target;
+
+/// (LLDB extension) If your target is using the `qRegisterInfo` packet, this
```
daniel5151, line 5: nit: similarly, since the module these types are contained within is marked as `(LLDB extension)`, there's no need to repeat that comment within the module.

src/target/ext/register_info_override.rs
```diff
@@ -0,0 +1,59 @@
...
+/// (LLDB extension) This struct is used internally by `gdbstub` to wrap a
+/// callback function which has to be used to direct the `qRegisterInfo` query.
+pub struct Callback<'a> {
+    /// The callback function that is directing the `qRegisterInfo` query.
+    pub(crate) cb: &'a mut dyn FnMut(Option<Register<'_>>),
+    /// A token to guarantee the callback has been used.
+    pub(crate) token: CallbackToken<'a>,
+}
+
+/// (LLDB extension) Trait which acts as a forwarder for the callback. Use the
```
daniel5151, line 19: nit: you can remove this `impl` block doc comment entirely

src/stub/core_impl/register_info.rs
```diff
@@ -0,0 +1,134 @@
...
+use crate::arch::Arch;
+
+impl<T: Target, C: Connection> GdbStubImpl<T, C> {
+    pub(crate) fn handle_register_info(
+        &mut self,
+        res: &mut ResponseWriter<'_, C>,
+        target: &mut T,
+        command: RegisterInfo,
+    ) -> Result<HandlerStatus, Error<T::Error, C::Error>> {
+        if !target.use_register_info() {
```
daniel5151, line 14: I've got a sneaking suspicion that this dead-code-elimination guard might need to be "beefed up" a bit to make sure we aren't introducing undue binary overhead to targets not using register info...

Can you double check and make sure that `./example_no_std/check_size.sh` returns the same value before/after this PR?

src/arch.rs
```diff
@@ -255,3 +282,121 @@ pub enum SingleStepGdbBehavior {
     #[doc(hidden)]
     Unknown,
 }
+
+/// (LLDB extension)
```
daniel5151, line 286:
```suggestion
/// LLDB-specific types supporting [`Arch::register_info`] and [`RegisterInfoOverride`](path::to::this::type) APIs
```

daniel5151 (changes requested):
mostly doc and style nits, with one somewhat pertinent question regarding possible binary overhead changes...

thanks for bearing with me through all these review cycles, I think we are almost done :)

src/stub/core_impl/register_info.rs
jawilk: Testing off commit 1fde6f7
before this PR
```
section               size      addr
...
.text                15010      1792
...
Total                19524
```
after this PR
```
section               size      addr
...
.text                14994      1792
...
Total                19508
```

daniel5151: huh, wasn't expecting a shrink, but hey, I'll take it! the LLVM optimizer works in mysterious ways...

daniel5151:
@jawilk, #106 introduced a couple of merge conflicts

jawilk:
> @jawilk, #106 introduced a couple of merge conflicts

done

daniel5151:
Hey, quick update - I've been surprisingly busy lately, so I'll probably get a chance to look at this again sometime this weekend. Sorry about that!

jawilk:
> Hey, quick update - I've been surprisingly busy lately, so I'll probably get a chance to look at this again sometime this weekend. Sorry about that!

no worries, thanks for the update!

src/target/ext/register_info_override.rs
```diff
@@ -0,0 +1,54 @@
...
+}
+
+/// Target Extension - Override the target register info
+/// specified by `Target::Arch`.
+///
+/// _Note:_ Unless you're working with a particularly dynamic,
+/// runtime-configurable target, it's unlikely that you'll need to implement
+/// this extension.
+pub trait RegisterInfoOverride: Target {
+    /// Invoke `reg_info.write(reg)` where `reg` is a [`Register`
```
daniel5151, line 42: nit: weird little formatting issue. reminds you why `wrap_comment` is still a nightly feature :)

daniel5151 (approved):
Only thing left are some minor comment formatting issues, but honestly, I'll just take care of those myself, since I'll be making some misc. commits prior to releasing 0.6.2 anyways...

Thanks again for the PR!

I'll try and push out 0.6.2 sometime in the next few days :)

daniel5151:
Oh, also, heads up: in 4a8f3258a04062ac9299b59b4eab8d934ad48713, I renamed all instances of `register_info` to `lldb_register_info`, to make it super extra clear that this is a LLDB specific extension, and if you only care about gdb, you don't need to implement it.

daniel5151:
@jawilk 0.6.2 has been released :)

daniel5151:
...oh, and thanks for the Coffee!

PR #106, opened 2022-06-08
Add support for qThreadExtraInfo

### LINKED ISSUE #104: `qThreadExtraInfo` support
While porting probe-rs to use gdbstub I found two packets that would be helpful to respond to:

~~`qC` - this queries the current thread ID.  WIthout responding GDB works correctly, but the client emits a message that may be confusing to end users.~~  Edit: `qC` is no longer needed after debugging an issue with `?` in the thread.

`qThreadExtraInfo` - this allows the stub to provide a string of extra data about a thread.  It's a free form string that is displayed on the client and can be useful for adding extra detailed status or thread names, for example.

Both of these would produce a better experience for the probe-rs logic and I'd be glad to take a shot at producing a PR that adds the logic.  Before I do I wanted some feedback on how to approach this.  

Both packets feel like appropriate functions on the `MultiThreadBase` trait.  To avoid a break I could add them with default implementations - `qC` could return the first value from `list_active_threads` and `qThreadExtraInfo` could return a 0 byte string by default.  This avoids a break, but it puts a bit more burden on the implementor to realize these are options for them to extend.

Any thoughts / feedback on this?

thefaxman:
### Description

Add a new sub-IDET for `MultiThreadBase` to respond with extra information about a thread.

Closes #104 

### API Stability

- [X] This PR does not require a breaking API change

### Checklist

- Implementation
  - [X] `cargo build` compiles without `errors` or `warnings`
  - [X] `cargo clippy` runs without `errors` or `warnings`
  - [X] `cargo fmt` was run
  - [X] All tests pass
- Documentation
  - [X] rustdoc + approprate inline code comments
  - [X] Updated CHANGELOG.md
  - [X] (if appropriate) Added feature to "Debugging Features" in README.md
  - [X] Included a basic sample implementation in `examples/armv4t`
  - [X] Included output of running `examples/armv4t` with `RUST_LOG=trace` + any relevant GDB output under the "Validation" section below
  - [X] Confirmed that IDET can be optimized away (using `./scripts/test_dead_code_elim.sh` and/or `./example_no_std/check_size.sh`)

### Validation

<details>
<summary>GDB output</summary>

```
(gdb) info thread
  Id   Target Id            Frame 
* 1    Thread 1.1 (CPU Cpu) 0x55550000 in ?? ()
  2    Thread 1.2 (CPU Cop) 0x55550000 in ?? ()
(gdb) 
```

</details>

<details>
<summary>armv4t output</summary>

```
 TRACE gdbstub::protocol::recv_packet     > <-- $qThreadExtraInfo,p1.1#85
 TRACE gdbstub::protocol::response_writer > --> $43505520437075#d2
 TRACE gdbstub::protocol::recv_packet     > <-- $qThreadExtraInfo,p1.2#86
 TRACE gdbstub::protocol::response_writer > --> $43505520436f70#02
```
</details>

### DIFF
--- a/examples/armv4t_multicore/gdb.rs
+++ b/examples/armv4t_multicore/gdb.rs
@@ -126,6 +126,13 @@ impl MultiThreadBase for Emu {
     ) -> Option<target::ext::base::multithread::MultiThreadResumeOps<'_, Self>> {
         Some(self)
     }
+
+    #[inline(always)]
+    fn support_thread_extra_info(
+        &mut self,
+    ) -> Option<gdbstub::target::ext::thread_extra_info::ThreadExtraInfoOps<'_, Self>> {
+        Some(self)
+    }
 }
 
 impl MultiThreadResume for Emu {
@@ -271,4 +278,21 @@ impl target::ext::breakpoints::HwWatchpoint for Emu {
 
         Ok(true)
     }
+}
+
+impl target::ext::thread_extra_info::ThreadExtraInfo for Emu {
+    fn thread_extra_info(&self, tid: Tid, buf: &mut [u8]) -> Result<usize, Self::Error> {
+        let cpu_id = tid_to_cpuid(tid)?;
+        let info = format!("CPU {:?}", cpu_id);
+
+        Ok(copy_to_buf(info.as_bytes(), buf))
+    }
+}
+
+/// Copy all bytes of `data` to `buf`.
+/// Return the size of data copied.
+pub fn copy_to_buf(data: &[u8], buf: &mut [u8]) -> usize {
+    let len = buf.len().min(data.len());
+    buf[..len].copy_from_slice(&data[..len]);
+    len
 }
--- a/src/protocol/commands/_qThreadExtraInfo.rs
+++ b/src/protocol/commands/_qThreadExtraInfo.rs
@@ -1,0 +1,32 @@
+use super::prelude::*;
+
+use crate::protocol::common::thread_id::ThreadId;
+use crate::protocol::ConcreteThreadId;
+
+#[derive(Debug)]
+pub struct qThreadExtraInfo<'a> {
+    pub id: ConcreteThreadId,
+
+    pub buf: &'a mut [u8],
+}
+
+impl<'a> ParseCommand<'a> for qThreadExtraInfo<'a> {
+    #[inline(always)]
+    fn from_packet(buf: PacketBuf<'a>) -> Option<Self> {
+        let (buf, body_range) = buf.into_raw_buf();
+        let body = buf.get(body_range.start..body_range.end)?;
+
+        if body.is_empty() {
+            return None;
+        }
+
+        match body {
+            [b',', body @ ..] => {
+                let id = ConcreteThreadId::try_from(ThreadId::try_from(body).ok()?).ok()?;
+
+                Some(qThreadExtraInfo { id, buf })
+            }
+            _ => None,
+        }
+    }
+}
--- a/src/stub/core_impl/thread_extra_info.rs
+++ b/src/stub/core_impl/thread_extra_info.rs
@@ -1,0 +1,37 @@
+use super::prelude::*;
+use crate::protocol::commands::ext::ThreadExtraInfo;
+use crate::target::ext::base::BaseOps;
+
+impl<T: Target, C: Connection> GdbStubImpl<T, C> {
+    pub(crate) fn handle_thread_extra_info<'a>(
+        &mut self,
+        res: &mut ResponseWriter<'_, C>,
+        target: &mut T,
+        command: ThreadExtraInfo<'a>,
+    ) -> Result<HandlerStatus, Error<T::Error, C::Error>> {
+        let ops = match target.base_ops() {
+            BaseOps::SingleThread(_) => return Ok(HandlerStatus::Handled),
+            BaseOps::MultiThread(ops) => match ops.support_thread_extra_info() {
+                Some(ops) => ops,
+                None => return Ok(HandlerStatus::Handled),
+            },
+        };
+
+        crate::__dead_code_marker!("thread_extra_info", "impl");
+
+        let handler_status = match command {
+            ThreadExtraInfo::qThreadExtraInfo(info) => {
+                let size = ops
+                    .thread_extra_info(info.id.tid, info.buf)
+                    .map_err(Error::TargetError)?;
+                let data = info.buf.get(..size).ok_or(Error::PacketBufferOverflow)?;
+
+                res.write_hex_buf(data)?;
+
+                HandlerStatus::Handled
+            }
+        };
+
+        Ok(handler_status)
+    }
+}
--- a/src/target/ext/thread_extra_info.rs
+++ b/src/target/ext/thread_extra_info.rs
@@ -1,0 +1,22 @@
+//! Provide extra information for a thread
+use crate::common::Tid;
+use crate::target::Target;
+
+/// Target Extension - Provide extra information for a thread
+pub trait ThreadExtraInfo: Target {
+    /// Provide extra information about a thread
+    ///
+    /// GDB queries for extra information for a thread as part of the
+    /// `info threads` command.  This function will be called once
+    /// for each active thread.
+    ///
+    /// A string can be copied into `buf` that will then be displayed
+    /// to the client.  The string is displayed as `(value)`, such as:
+    ///
+    /// `Thread 1.1 (value)`
+    ///
+    /// Return the number of bytes written into `buf`.
+    fn thread_extra_info(&self, tid: Tid, buf: &mut [u8]) -> Result<usize, Self::Error>;
+}
+
+define_ext!(ThreadExtraInfoOps, ThreadExtraInfo);
--- a/README.md
+++ b/README.md
@@ -87,6 +87,7 @@ Of course, most use-cases will want to support additional debugging features as 
     -   Access the remote target's filesystem to read/write file
     -   Can be used to automatically read the remote executable on attach (using `ExecFile`)
 -   Read auxiliary vector (`info auxv`)
+-   Extra thread info (`info threads`)
 
 _Note:_ GDB features are implemented on an as-needed basis by `gdbstub`'s contributors. If there's a missing GDB feature that you'd like `gdbstub` to implement, please file an issue and/or open a PR!
 
--- a/src/protocol/commands.rs
+++ b/src/protocol/commands.rs
@@ -92,6 +92,7 @@ macro_rules! commands {
                     fn support_reverse_step(&mut self) -> Option<()>;
                     fn support_reverse_cont(&mut self) -> Option<()>;
                     fn support_x_upcase_packet(&mut self) -> Option<()>;
+                    fn support_thread_extra_info(&mut self) -> Option<()>;
                 }
 
                 impl<T: Target> Hack for T {
@@ -144,6 +145,14 @@ macro_rules! commands {
                             Some(())
                         } else {
                             None
+                        }
+                    }
+
+                    fn support_thread_extra_info(&mut self) -> Option<()> {
+                        use crate::target::ext::base::BaseOps;
+                        match self.base_ops() {
+                            BaseOps::SingleThread(_) => None,
+                            BaseOps::MultiThread(ops) => ops.support_thread_extra_info().map(drop),
                         }
                     }
                 }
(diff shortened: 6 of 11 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@ All notable changes to this project will be documented in this file.
 
 This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+# 0.6.2
```
daniel5151, line 5: you can omit this changelog update. I realized that I usually end up writing these myself anyways, so I'll be removing that step from the PR checklist shortly...

examples/armv4t_multicore/gdb.rs
```diff
@@ -272,3 +280,20 @@ impl target::ext::breakpoints::HwWatchpoint for Emu {
...
+        let cpu_id = tid_to_cpuid(tid).map_err(TargetError::Fatal)?;
+        let info = format!("CPU {:?}", cpu_id);
+
+        Ok(copy_to_buf(info.as_bytes(), buf))
+    }
+}
+
+/// Copy all bytes of `data` to `buf`.
+/// Return the size of data copied.
+pub fn copy_to_buf(data: &[u8], buf: &mut [u8]) -> usize {
```
daniel5151, line 295: I believe this function should already be floating around somewhere in armv4t... no need to redefine it.

examples/armv4t_multicore/gdb.rs
```diff
@@ -4,6 +4,7 @@ use gdbstub::common::{Signal, Tid};
 use gdbstub::target;
 use gdbstub::target::ext::base::multithread::{MultiThreadBase, MultiThreadResume};
 use gdbstub::target::ext::breakpoints::WatchKind;
+use gdbstub::target::ext::thread_extra_info::ThreadExtraInfo;
```
daniel5151, line 7: for consistency, could you use the fully qualified path in the impl block, rather than importing it?

src/protocol/commands/_qThreadExtraInfo.rs
```diff
@@ -0,0 +1,31 @@
...
+    pub id: SpecificThreadId,
+
+    pub buf: &'a mut [u8],
+}
+
+impl<'a> ParseCommand<'a> for qThreadExtraInfo<'a> {
+    #[inline(always)]
+    fn from_packet(buf: PacketBuf<'a>) -> Option<Self> {
+        let (buf, body_range) = buf.into_raw_buf();
+        let (body, buf) = buf[body_range.start..].split_at_mut(body_range.end - body_range.start);
```
daniel5151, line 17: take a look at how “vFilePread` is implemented. 

In this case, since you don't need to hold on to any references into the `buf` after parsing is complete, you are able to reuse the entirety of the buffer.

src/stub/core_impl/thread_extra_info.rs
```diff
@@ -0,0 +1,37 @@
...
+            BaseOps::SingleThread(_) => return Ok(HandlerStatus::Handled),
+            BaseOps::MultiThread(ops) => ops,
+        };
+
+        let handler_status = match command {
+            ThreadExtraInfo::qThreadExtraInfo(info) => {
+                if let Some(ops) = ops.support_thread_extra_info() {
+                    if let SpecificIdKind::WithId(tid) = info.id.tid {
+                        let size = ops.extra_info(tid, info.buf).handle_error()?;
+                        let data = &mut info.buf[..size];
```
daniel5151, line 23: take a look at how `vFilePread` handles this. `gdbstub` takes care to avoid introducing any panicking paths, which includes those that might occur from array indexing.

src/target/ext/thread_extra_info.rs
```diff
@@ -0,0 +1,22 @@
...
+    /// `info threads` command.  This function will be called once
+    /// for each active thread.
+    ///
+    /// A string can be copied into `buf` that will then be displayed
+    /// to the client.  The string is displayed as `(value)`, such as:
+    ///
+    /// `Thread 1.1 (value)`
+    ///
+    /// Return the number of bytes written into `buf`.
+    fn extra_info(&self, tid: Tid, buf: &mut [u8]) -> TargetResult<usize, Self>;
```
daniel5151, line 19: nit: `thread_extra_info`?

src/stub/core_impl/thread_extra_info.rs
```diff
@@ -0,0 +1,37 @@
...
+    pub(crate) fn handle_thread_extra_info<'a>(
+        &mut self,
+        res: &mut ResponseWriter<'_, C>,
+        target: &mut T,
+        command: ThreadExtraInfo<'a>,
+    ) -> Result<HandlerStatus, Error<T::Error, C::Error>> {
+        let ops = match target.base_ops() {
+            BaseOps::SingleThread(_) => return Ok(HandlerStatus::Handled),
+            BaseOps::MultiThread(ops) => ops,
+        };
```
daniel5151, line 16: so, this guard need to be modified to ensure that the rest of the function gets optimized out if `support_thread_extra_info` returns `None`.

The following should do what you want (along with simplifying some of the `ops` code down below as well, removing the need for that `PacketUnexpected` arm).

```suggestion
        let ops = match target.base_ops() {
            BaseOps::SingleThread(_) => return Ok(HandlerStatus::Handled),
            BaseOps::MultiThread(ops) => match ops.support_thread_extra_info() {
                Some(ops) => ops,
                None => return Ok(HandlerStatus::Handled),
            },
        };
```

And please add in the `__dead_code_marker` as present in other handlers, and validate that this bit of code does indeed get dead code eliminated in the case that the target doesn't implement this IDET. (i.e: by temporarily commenting it out in the armv4t example, and running the `test_dead_code_elim.sh` script). 

In addition, please make sure that running `./example_no_std/check_size.sh` returns the same value before / after this PR.

src/target/ext/thread_extra_info.rs
daniel5151, line 19: reading through the docs, it seems that `qThreadExtraInfo` is _not_ allowed to return a arbitrary error value. As such, this signature should be modified to return a regular `Result<usize, Self::Error>`.

This is a common inconsistency found in the GDB RSP, where only some methods support reporting error codes, but not others. It's pretty annoying...

daniel5151 (changes requested):
All in all, this is looking good!

CHANGELOG.md
thefaxman: Reverted

examples/armv4t_multicore/gdb.rs
thefaxman: Done

src/protocol/commands/_qThreadExtraInfo.rs
thefaxman: Thanks for the pointer, I've matched the logic

src/stub/core_impl/thread_extra_info.rs
thefaxman: Updated

src/target/ext/thread_extra_info.rs
thefaxman: Updated

src/stub/core_impl/thread_extra_info.rs
thefaxman: Updated as requested.

`test_dead_code_elim.sh` shows the path is getting eliminated properly.  `check_size.sh` shows an increase of exactly 1 byte in .text.  I'm still trying to trace down where that is coming from.

src/target/ext/thread_extra_info.rs
thefaxman: Yep, you're right.  It's clear that the GDB protocol has evolved very organically so there's a lot of pitfalls.

Updated.

src/stub/core_impl/thread_extra_info.rs
daniel5151: huh, interesting. a single byte difference? that's weird...

well, in any case, I wouldn't worry too much about that 😄

examples/armv4t_multicore/gdb.rs
daniel5151: ahh, indeed: examples/armv4t/gdb/mod.rs:39

src/stub/core_impl/thread_extra_info.rs
thefaxman: Looks like the 1 byte difference is just the optimizer reordering some jmp / calls such that we end up with a slightly different generated main code.  None of the calls in question are in the new IDET paths, but it's enough to cause the compiler to generate code slightly differently.

I can provide assembly diffs if interested but my read is that this may be very difficult to avoid.

daniel5151: yep, that's what I would've guessed. I wouldn't sweat it - different rustc versions introduce plenty more variation already. the key thing is that we aren't seeing massive bumps in size (i.e: hundreds of bytes).

examples/armv4t_multicore/gdb.rs
thefaxman: There was a copy in armv4t, but not armv4t_multicore

daniel5151: ahh, didn't realize this was `armv4t_multicore`, oops 😅 
All good

src/target/ext/thread_extra_info.rs
thefaxman: I missed part of this in the second push, should all be fixed to use `Result<usize, Self::Error>` now.

thefaxman:
I think I've addressed all the feedback - let me know if there's any additional items you'd like to see addressed

bet4it (changes requested):
Refer: https://github.com/daniel5151/gdbstub/blob/master/src/protocol/commands/_p.rs

src/stub/core_impl/thread_extra_info.rs
```diff
@@ -0,0 +1,40 @@
...
+        let handler_status = match command {
+            ThreadExtraInfo::qThreadExtraInfo(info) => {
+                if let SpecificIdKind::WithId(tid) = info.id.tid {
+                    let size = ops
+                        .thread_extra_info(tid, info.buf)
+                        .map_err(Error::TargetError)?;
+                    let data = info.buf.get(..size).ok_or(Error::PacketBufferOverflow)?;
+
+                    res.write_hex_buf(data)?;
+                }
```
daniel5151, line 32: You'll want to add an extra else arm here:

```rust
} else {
    return Err(Error::PacketParse(
        crate::protocol::PacketParseError::MalformedCommand,
    ))
}
```

The _best_ solution would actually be to go the extra mile and introduce a new `ThreadId` variant at the `protocol` level that doesn't support the `All` variant. Something like `ConcreteThreadId` (not a great name, but whatever). That way, an invalid thread-id would be caught at the packet parse level, rather than within the handler.

It would be nice to implement that latter solution, but it's fine if you'd like to stick with the quick-and-dirty fix I proposed above.

src/protocol/commands/_qThreadExtraInfo.rs
```diff
@@ -0,0 +1,35 @@
...
+        let body = buf.get_mut(body_range.start..body_range.end)?;
+
+        if body.is_empty() {
+            return None;
+        }
+
+        match body {
+            [b',', body @ ..] => {
+                let mut body = body.splitn_mut_no_panic(3, |b| *b == b',');
+                let id = SpecificThreadId::try_from(ThreadId::try_from(body.next()?).ok()?).ok()?;
```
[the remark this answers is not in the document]
thefaxman: Updated

src/protocol/commands/_qThreadExtraInfo.rs
```diff
@@ -0,0 +1,35 @@
...
+    pub id: SpecificThreadId,
+
+    pub buf: &'a mut [u8],
+}
+
+impl<'a> ParseCommand<'a> for qThreadExtraInfo<'a> {
+    #[inline(always)]
+    fn from_packet(buf: PacketBuf<'a>) -> Option<Self> {
+        let (buf, body_range) = buf.into_raw_buf();
+        let body = buf.get_mut(body_range.start..body_range.end)?;
```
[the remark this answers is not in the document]
thefaxman: Updated

daniel5151 (changes requested):
Looks good.

It's a bit late for me here, so I'll give it one last once-over in the morning just in case, but it should be good to go after resolving this last thing.

src/stub/core_impl/thread_extra_info.rs
thefaxman: I like that - I added a `ConcreteThreadId` type.  Naming is hard - I couldn't come up with another alternative that I felt expressed things more clearly.

As suggested this simplified the handler logic.

daniel5151:
Thanks again for the contribution! 

Once #103 lands, I'll push out `0.6.2` that includes these and other changes.

daniel5151:
@thefaxman 0.6.2 has been released :)

PR #107, opened 2022-07-20
SingleRegisterAccess: Support unavailable regs

ptosi:
### Description

Make gdbstub report to the GDB client the state of an unavailable (but valid) register by following the [registers packet protocol](https://sourceware.org/gdb/onlinedocs/gdb/Packets.html#read-registers-packet):

> [...] the stub may also return a string of literal `x`s in place of
> the register data digits, to indicate that the corresponding register
> has not been collected, thus its value is unavailable.

### API Stability

- [x] This PR does not require a breaking API change but it would probably benefit from introducing an `Option` instead of the C-ish approach of giving special meaning to `0`

### Checklist

- Documentation
  - [x] Ensured any public-facing `rustdoc` formatting looks good (via `cargo doc`)
  - [x] (if appropriate) Added feature to "Debugging Features" in README.md
- Validation
  - [x] Included output of running `examples/armv4t` with `RUST_LOG=trace` + any relevant GDB output under the "Validation" section below
  - [x] Included output of running `./example_no_std/check_size.sh` before/after changes under the "Validation" section below

### Validation

<details>
<summary>GDB output</summary>

```
(gdb) info registers SPSR_EL3
SPSR_EL3       <unavailable>
```

</details>

<details>
<summary>examples/armv4t</summary>

```
$ RUST_LOG=trace cargo run --example armv4t
   Compiling gdbstub v0.6.2 (/usr/local/google/home/ptosi/src/gdbstub)
    Finished dev [unoptimized + debuginfo] target(s) in 1.76s
     Running `target/debug/examples/armv4t`
loading section ".text" into memory from [0x55550000..0x55550078]
Setting PC to 0x55550000
Waiting for a GDB connection on "127.0.0.1:9001"...
```
then
```
(gdb) target remote localhost:9001
```
makes the GDB client learn about the `Unavailable` register through the XML:
```
Debugger connected from 127.0.0.1:33718
 TRACE gdbstub::protocol::recv_packet > <-- +
 TRACE gdbstub::protocol::recv_packet > <-- $qSupported:multiprocess+;swbreak+;hwbreak+;qRelocInsn+;fork-events+;vfork-events+;exec-events+;vContSupported+;QThreadEvents+;no-resumed+;xmlRegisters=i386#6a
[...]

PR #108, opened 2022-07-21
check_size.sh: Allow running script from other dirs

ptosi:
### Description

As shown in https://github.com/daniel5151/gdbstub/pull/107#issuecomment-1191802780, it was easy to misinterpret the instructions as running the script from the root of the directory (which would break). Make it possible to do so.


daniel5151 (approved):
Thanks!

Issue #112, opened 2022-10-19
`TargetMismatch` did not help me fix wrong `Arch` implementation


geigerzaehler:
I recently started experimenting with this crate for AVR. When implementing the target I ran into `TargetMismatch` errors . The information in those errors was not enough for me to fix the issues—I had to dig through the source code instead.

The first issue was that `Arch::Usize` was different from what GDB send. To fix the issue I needed to know what GDB expected. The second issue was that `Arch::BreakpointKind` kind was different from what GDB sent. (I assumed that it would be `()`).

What would have helped me to fix this issues faster is an error message pointing to what to fix and how to fix it.

(Happy to provide a PR for this if you want me to.)

daniel5151:
Taking a step back: I think the broader issue here is that there is plenty of room for improvement in `gdbstub`'s error types _in general_.

I've learned quite a bit about proper error handling in Rust in the years since starting `gdbstub`, and one thing that's for certain is that you want to have error types that uniquely identify where in the code something failed - something that most certainly isn't the case in the library today.

I spent a bit of time looking through the current code, and honestly - I'm not sure if I have a good "easy" solution off the top of my head. My gut feeling is that the current approach of a single top-level Error enum isn't ideal (it has some semver compat guarantees I'm not a fan of, and isn't particularly well "organized" wrt. different classes of error), and that a more holistic rework of the error type is in order.

That is to say: thanks for the feedback, and for offering to help out with a PR here, but I think this is something I should take a stab at myself at some point :)

PR #115, opened 2022-10-27
`read_addrs()` can return fewer bytes than requested

### LINKED ISSUE #111: Allow `read_addrs()` to send fewer bytes than requested
GDBs `m` command allows the stub to respond with fewer bytes than requested if part of the memory is not accessible. Implementations of `SingleThreadBase::read_addrs()` cannot make use of that feature. If an implementation writes fewer bytes to the `data` argument the remaining bytes will be zero.

(Happy to provide a PR for this if you want me to.)

geigerzaehler:
### Description

The `read_addrs()` method for targets now returns a number of bytes that were read. This allows a target to gracefully signal the client that memory at an address range is not accessible.

I’ve implemented this functionality for the armv4t example to be able to test it.

Closes #111.

### API Stability

This PR requires a breaking API change.

### Checklist

- Implementation
  - [x] `cargo fmt` was run
  - [x] All tests pass
- Documentation
  - [ ] Updated CHANGELOG.md
  - ~~(if appropriate) Added feature to "Debugging Features" in README.md~~
- ~~_If implementing a new protocol extension IDET_~~
- ~~_If upstreaming an `Arch` implementation_~~

I’m not sure how to update the changelog.

#### Read before accessible area

```
(gdb) x 0x00
0x0:    Cannot access memory at address 0x0
```

```
TRACE gdbstub::protocol::recv_packet     > <-- $m0,4#fd
TRACE gdbstub::protocol::response_writer > --> $#00
```

#### Read partially accessible memory

```
(gdb) x/4xb 0x6554fffe
0x6554fffe:     0x00    0x00    Cannot access memory at address 0x65550000
```

```
TRACE gdbstub::protocol::recv_packet     > <-- $m6554fffe,1#35
TRACE gdbstub::protocol::response_writer > --> $00#60
TRACE gdbstub::protocol::recv_packet     > <-- $m6554ffff,1#36
TRACE gdbstub::protocol::response_writer > --> $00#60
TRACE gdbstub::protocol::recv_packet     > <-- $m65550000,1#5f
TRACE gdbstub::protocol::response_writer > --> $#00
```

#### Read after accessible area

```
(gdb) x 0x6555000e
0x6555000e:     Cannot access memory at address 0x6555000e
```

```
TRACE gdbstub::protocol::recv_packet     > <-- $m6555000e,4#97
TRACE gdbstub::protocol::response_writer > --> $#00
```

### DIFF
--- a/examples/armv4t/gdb/mod.rs
+++ b/examples/armv4t/gdb/mod.rs
@@ -196,11 +196,22 @@ impl SingleThreadBase for Emu {
         Some(self)
     }
 
-    fn read_addrs(&mut self, start_addr: u32, data: &mut [u8]) -> TargetResult<(), Self> {
-        for (addr, val) in (start_addr..).zip(data.iter_mut()) {
+    fn read_addrs(&mut self, start_addr: u32, data: &mut [u8]) -> TargetResult<usize, Self> {
+        // These values are taken from the link script.
+        const MEMORY_ORIGIN: u32 = 0x5555_0000;
+        const MEMORY_LENGTH: u32 = 0x1000_0000;
+        const MEMORY_END: u32 = MEMORY_ORIGIN + MEMORY_LENGTH;
+
+        if !(MEMORY_ORIGIN..MEMORY_END).contains(&start_addr) {
+            return Err(TargetError::NonFatal);
+        }
+
+        let end_addr = std::cmp::min(start_addr + data.len() as u32, MEMORY_END);
+
+        for (addr, val) in (start_addr..end_addr).zip(data.iter_mut()) {
             *val = self.mem.r8(addr)
         }
-        Ok(())
+        Ok((end_addr - start_addr) as usize)
     }
 
     fn write_addrs(&mut self, start_addr: u32, data: &[u8]) -> TargetResult<(), Self> {
--- a/src/stub/core_impl/base.rs
+++ b/src/stub/core_impl/base.rs
@@ -238,7 +238,7 @@ impl<T: Target, C: Connection> GdbStubImpl<T, C> {
 
                     let addr = addr + NumCast::from(i).ok_or(Error::TargetMismatch)?;
                     let data = &mut buf[..chunk_size];
-                    match target.base_ops() {
+                    let data_len = match target.base_ops() {
                         BaseOps::SingleThread(ops) => ops.read_addrs(addr, data),
                         BaseOps::MultiThread(ops) => {
                             ops.read_addrs(addr, data, self.current_mem_tid)
@@ -249,6 +249,8 @@ impl<T: Target, C: Connection> GdbStubImpl<T, C> {
                     n -= chunk_size;
                     i += chunk_size;
 
+                    // TODO: add more specific error variant?
+                    let data = data.get(..data_len).ok_or(Error::PacketBufferOverflow)?;
                     res.write_hex_buf(data)?;
                 }
                 HandlerStatus::Handled
--- a/src/target/ext/base/multithread.rs
+++ b/src/target/ext/base/multithread.rs
@@ -43,17 +43,25 @@ pub trait MultiThreadBase: Target {
         None
     }
 
-    /// Read bytes from the specified address range.
+    /// Read bytes from the specified address range and return the number of
+    /// bytes that were read.
     ///
-    /// If the requested address range could not be accessed (e.g: due to
-    /// MMU protection, unhanded page fault, etc...), an appropriate non-fatal
-    /// error should be returned.
+    /// Implementations may return a number `n` that is less than `data.len()`
+    /// to indicate that memory starting at `start_addr + n` cannot be
+    /// accessed.
+    ///
+    /// Implemenations may also return an appropriate non-fatal error if the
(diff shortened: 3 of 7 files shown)


examples/armv4t/gdb/mod.rs
```diff
@@ -196,11 +196,22 @@ impl SingleThreadBase for Emu {
...
-    fn read_addrs(&mut self, start_addr: u32, data: &mut [u8]) -> TargetResult<(), Self> {
-        for (addr, val) in (start_addr..).zip(data.iter_mut()) {
+    fn read_addrs(&mut self, start_addr: u32, data: &mut [u8]) -> TargetResult<usize, Self> {
+        // These values are taken from the link script.
+        const MEMORY_ORIGIN: u32 = 0x5555_0000;
+        const MEMORY_LENGTH: u32 = 0x1000_0000;
+        const MEMORY_END: u32 = MEMORY_ORIGIN + MEMORY_LENGTH;
+
+        if !(MEMORY_ORIGIN..MEMORY_END).contains(&start_addr) {
+            return Ok(0);
```
daniel5151, line 206: It feels like it'd be more appropriate to return an error here? I know that's not what the old implementation did, but if we're already here changing things up...

src/stub/core_impl/base.rs
```diff
@@ -238,18 +238,19 @@ impl<T: Target, C: Connection> GdbStubImpl<T, C> {
...
                         }
                     }
                     .handle_error()?;
+                    debug_assert!(data_len <= data.len());
 
                     n -= chunk_size;
                     i += chunk_size;
 
-                    res.write_hex_buf(data)?;
+                    res.write_hex_buf(&data[..data_len])?;
```
daniel5151, line 253: one key goal of `gdbstub` is to be panic free, which means this code needs to be resilient against misbehaving target implementations that return a longer than expected length.

that means no asserts/debug_asserts, and no direct indexing (unless the bounds check has been optimized out)

The way this is currently done throughout the codebase is by doing something like

```rust
// TODO: add more specific error variant?
data.get(..data_len).ok_or(Error::PacketBufferOverflow)?
```

If you grep for that error type around the codebase, you'll find many comments along the lines of `// TODO: add more specific error variant?`, since it's not _really_ appropriate to recycle this error for this use case.

...buuuut that's a bigger fish to fry. for this PR, just do what other code is doing (including that same comment), and I might take a pass at fixing it as part of tackling #112

daniel5151 (changes requested):
all up this looks good!

oh, and I suppose I don't actually mention it anywhere (I really should): I use nightly rustfmt in `gdbstub` to automatically wrap doc comments, so you'll have to fix that.

geigerzaehler:
I’ve addressed all your comments. Sorry I took a while.

daniel5151:
> Sorry I took a while.

all good :)

I'll take a look in the next day or two

src/target/ext/base/singlethread.rs
```diff
@@ -32,15 +32,19 @@ pub trait SingleThreadBase: Target {
...
     ///
-    /// Implementations may return a number `n` that is less than `data.len()` to indicate that
-    /// memory starting at `start_addr + n` cannot be accessed.
+    /// Implementations may return a number `n` that is less than `data.len()`
+    /// to indicate that memory starting at `start_addr + n` cannot be
+    /// accessed.
     ///
-    /// Implemenations may also return an appropriate non-fatal if the requested address range
-    /// could not be accessed (e.g: due to MMU protection, unhanded page fault, etc...).
+    /// Implemenations may also return an appropriate non-fatal if the requested
```
daniel5151, line 42: missed a word. ditto in the same comment up in `multithread.rs`

```suggestion
    /// Implemenations may also return an appropriate non-fatal error if the requested
```

daniel5151 (changes requested):
one minor doc issue + a clippy lint that needs to get resolved, but aside from that, this LGTM!

geigerzaehler:
I‘ve fixed the lint and the docs. Should be good to go know :crossed_fingers:

src/target/ext/base/multithread.rs
```diff
@@ -43,17 +43,25 @@ pub trait MultiThreadBase: Target {
...
+    /// bytes that were read.
     ///
-    /// If the requested address range could not be accessed (e.g: due to
-    /// MMU protection, unhanded page fault, etc...), an appropriate non-fatal
-    /// error should be returned.
+    /// Implementations may return a number `n` that is less than `data.len()`
+    /// to indicate that memory starting at `start_addr + n` cannot be
+    /// accessed.
+    ///
+    /// Implemenations may also return an appropriate non-fatal if the requested
```
daniel5151, line 53:
```suggestion
    /// Implemenations may also return an appropriate non-fatal error if the requested
```

daniel5151 (changes requested):
...sorry to keep nitpicking, but in my earlier comment, I also mentioned that you needed to fix both places that had the doc issue.

src/target/ext/base/multithread.rs
geigerzaehler: My bad. I fixed the second instance too

PR #127, opened 2023-03-13
core_impl: fix support for no threads

### LINKED ISSUE #122: NoActiveThreads error when there are no active thread
When no threads are reported as part of `list_active_threads()`, gdbstub throws an error `gdbstub error during idle operation: NoActiveThreads`. A comment notes that this shouldn't happen, but it appears as though it does.

## Background

I've recently picked up gdbstub again and am working to integrate it into Xous more fully. It really is a fantastic project.

I have things mostly working now, and I'm filing off some of the rough edges. One of the rough edges is process selection. I have a `monitor` command to allow switching between processes, but the question arises of what process to connect to initially.

I support having `None` process, which is what I'd like to have selected initially. This coincides with a system that's fully running and does not halt when a debugger is attached.

However, returning early in `list_active_threads()` before adding any threads appears to result in this error.

xobs:
Sometimes, a target has no threads. This may be the case when a process is not yet selected, or when all threads have terminated in a specific process.

Support `None` as a return from `get_sane_any_tid()` and return an error to GDB. This addresses a `NoActiveThreads` error, and closes #122.


src/stub/core_impl/base.rs
```diff
@@ -302,17 +306,26 @@ impl<T: Target, C: Connection> GdbStubImpl<T, C> {
                 use crate::protocol::commands::_h_upcase::Op;
                 match cmd.kind {
                     Op::Other => match cmd.thread.tid {
-                        IdKind::Any => self.current_mem_tid = self.get_sane_any_tid(target)?,
+                        IdKind::Any => match self.get_sane_any_tid(target)? {
+                            Some(tid) => self.current_mem_tid = tid,
+                            None => {
+                                res.write_str("E01")?;
+                                return Ok(HandlerStatus::Handled);
+                            }
```
daniel5151, line 314: I actually have a bit of plumbing in place to streamline returning `Exx`-style errors:

```suggestion
                            None => return Err(Error::NonFatalError(1))
```

It's both more concise than invoking write_str + returning early, and also results in some slightly better codegen (wrt. binary size).

Can you switch both blocks over?

daniel5151 (changes requested):
Aside from that one comment that needs addressing, this LGTM

src/stub/core_impl/base.rs
xobs: Sure, I've made this change and pushed the result.

xobs:
Alright, I also marked `NoActiveThreads` as `#[deprecated]`

xobs:
It's been confusing because I need both PRs on my branch in order to do development, even though they are necessarily different features.

I think I've got it sorted now and it should be all set for merging.

daniel5151 (approved):
Nice, LGTM!

PR #154, opened 2024-10-11
feature to use core::error::Error instead of std::error::Error

### LINKED ISSUE #153: Supports core::error::Error for Rust 1.81+ as a bound on gdbstub::target::Target::Error
My application rely on `core::error::Error::source` to propagate the errors from the lower-level to the top-level. It would be great if `gdbstub` provide a support this!

ultimaweapon:
### Description

Not sure if the validation outputs are required because it is a small change.

Closes #153.

### API Stability

### Checklist

- Documentation
  - [x] Ensured any public-facing `rustdoc` formatting looks good (via `cargo doc`)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,7 @@ alloc = ["managed/alloc"]
 std = ["alloc"]
 trace-pkt = ["alloc"]
 paranoid_unsafe = []
+core_error = []
 
 # INTERNAL: enables the `__dead_code_marker!` macro.
 # used as part of the `scripts/test_dead_code_elim.sh`
--- a/README.md
+++ b/README.md
@@ -120,6 +120,8 @@ When using `gdbstub` in `#![no_std]` contexts, make sure to set `default-feature
     -   Add a `TargetError::Io` variant to simplify `std::io::Error` handling from Target methods.
 -   `paranoid_unsafe`
     -   Please refer to the [`unsafe` in `gdbstub`](#unsafe-in-gdbstub) section below for more details.
+-   `core_error`
+    -   Make `GdbStubError` implement [`core::error::Error`](https://doc.rust-lang.org/core/error/trait.Error.html) instead of `std::error::Error`.
 
 ## Examples
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -27,6 +27,9 @@
 //! - `paranoid_unsafe`
 //!     - Please refer to the [`unsafe` in `gdbstub`](https://github.com/daniel5151/gdbstub#unsafe-in-gdbstub)
 //!       section of the README.md for more details.
+//! - `core_error`
+//!     - Make `GdbStubError` implement [`core::error::Error`](https://doc.rust-lang.org/core/error/trait.Error.html)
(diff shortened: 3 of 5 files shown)


daniel5151:
Good start!

There are two other change you'll need to make before we can land this:

- Also update [`GdbStubError`](https://docs.rs/gdbstub/latest/gdbstub/stub/struct.GdbStubError.html) to impl `core::error::Error` instead of `std::error::Error` when this feature is enabled. Should be a similarly easy fix.
- Add a brief description of this new feature-flag to the docs in both [`README.md`](https://docs.rs/gdbstub/latest/gdbstub/#feature-flags), as well as the doc-comment in [`lib.rs`](https://docs.rs/gdbstub/latest/gdbstub/#feature-flags)

* * *

And yeah, feel free to delete the Validation section from the template - no need for a PR like this.

daniel5151:
And actually, now that I think about it - do we actually need a `core::error::Error` bound on the `type Error`?

I believe the downstream `Debug + Display` bounds on the various `impl` blocks should be giving users the flexibility to choose if they want to have their error type impl those various traits, without strictly requiring them to.

And thanks to the lack of type-erasure in gdbstub's API - if your error type implements `core::error::Error`, you'll still be able to leverage that implementation by using [`GdbStubError::into_target_error`](https://docs.rs/gdbstub/latest/gdbstub/stub/struct.GdbStubError.html#method.into_target_error), which returns the concrete type you specified.

ultimaweapon:
> And actually, now that I think about it - do we actually need a `core::error::Error` bound on the `type Error`?
> 
> I believe the downstream `Debug + Display` bounds on the various `impl` blocks should be giving users the flexibility to choose if they want to have their error type impl those various traits, without strictly requiring them to.
> 
> And thanks to the lack of type-erasure in gdbstub's API - if your error type implements `core::error::Error`, you'll still be able to leverage that implementation by using [`GdbStubError::into_target_error`](https://docs.rs/gdbstub/latest/gdbstub/stub/struct.GdbStubError.html#method.into_target_error), which returns the concrete type you specified.

You're right. I'll revert this changes and convert `std::error::Error` to `core::error::Error` instead.

ultimaweapon:
@daniel5151 do you have any preference how to choose between `std` and `core` version? I'm thinking to create a type alias but not sure if you would like to go this way. If yes let me know its name you like.

daniel5151:
Apologies for the delay - we just open-sourced https://github.com/microsoft/OpenVMM at $work, so my attention has been elsewhere the past few days 😅

I think a type-alias based approach is good? i.e: having a single `impl CoreError for Foo` block, where a cfg block determines what type backs the `type CoreError`?

daniel5151:
This approach looks good 👌

Two things before we can merge:
1. Address the `rustfmt` related style issue
2. Like I mentioned in https://github.com/daniel5151/gdbstub/pull/154#issuecomment-2407980989, make sure to update README.md and the doc-comment in lib.rs to mention this new feature.

daniel5151:
Awesome. Thanks again for sending in the PR! I'll merge this in, and cut a new gdbstub release with this new feature.

PR #179, opened 2025-12-27
MultiThreadSchedulerLocking support #178

Satar07:
### Description

This PR implements a new Protocol Extension IDET: `MultiThreadSchedulerLocking`.

Closes #178

#### Changes

- Introduced `MultiThreadSchedulerLocking` IDET under `MultiThreadResume`.
- Updated `do_vcont_multi_thread` to track the presence of wildcard/default continue actions.
- If no wildcard continue is found (indicating GDB expects scheduler locking), `gdbstub` now invokes `set_resume_action_scheduler_lock`.
- If the target does not implement this IDET but GDB requests locking, `gdbstub` returns a fatal `PacketUnexpected` error to avoid silent race conditions (as discussed in issue #178).
- Updated the `armv4t_multicore` example to demonstrate a working scheduler-locking implementation.

### API Stability

- \[x\] This PR does not require a breaking API change.

### Checklist

- Documentation
	- \[x\] Ensured any public-facing `rustdoc` formatting looks good
- Validation
	- \[x\] Included output of running `examples/armv4t_multicore`
	- \[x\] Included output of running `./example_no_std/check_size.sh`
- *If implementing a new protocol extension IDET*
	- \[x\] Included a basic sample implementation in `examples/armv4t_multicore`
	- \[x\] IDET can be optimized out (confirmed via `./example_no_std/check_size.sh`)

---

#### 1\. Functional Validation (`armv4t_multicore`)

Tested with GDB 15.0. Verified that when `set scheduler-locking on` is used, only the selected thread steps, while the other remain frozen at its last PC.

<details> <summary>GDB Session Output</summary>

Code snippet

```
0x55550000 in ?? ()
(gdb) info threads
  Id   Target Id            Frame
* 1    Thread 1.1 (CPU Cpu) 0x55550000 in ?? ()
  2    Thread 1.2 (CPU Cop) 0x55550000 in ?? ()
(gdb) b*0x55550000+0x30
Breakpoint 1 at 0x55550030
(gdb) c
Continuing.

Thread 1 hit Breakpoint 1, 0x55550030 in ?? ()
(gdb) info threads
  Id   Target Id            Frame
* 1    Thread 1.1 (CPU Cpu) 0x55550030 in ?? ()
  2    Thread 1.2 (CPU Cop) 0x55550044 in ?? ()
(gdb) set scheduler-locking on
(gdb) x/10i $pc
=> 0x55550030:  ldr     r3, [r11, #-16]
   0x55550034:  cmp     r3, #0
   0x55550038:  beq     0x55550030
   0x5555003c:  ldr     r3, [r11, #-8]
   0x55550040:  b       0x55550080
   0x55550044:  mov     r3, #0
   0x55550048:  str     r3, [r11, #-12]
   0x5555004c:  b       0x55550068
   0x55550050:  ldr     r3, [r11, #-8]
   0x55550054:  add     r3, r3, #1
(gdb) b*0x55550038
Breakpoint 3 at 0x55550038
(gdb) c
Continuing.

Thread 1 hit Breakpoint 3, 0x55550038 in ?? ()
(gdb) info threads
  Id   Target Id            Frame
* 1    Thread 1.1 (CPU Cpu) 0x55550038 in ?? ()
  2    Thread 1.2 (CPU Cop) 0x55550044 in ?? ()
```

</details>

#### 2\. Binary Size Check (`./example_no_std/check_size.sh`)

| Section | master | fix/resume\_multi | Diff |
| --- | --- | --- | --- |
| .text | 14128 | 14156 | **+28 bytes** |
| Total | 18586 | 18614 | **+28 bytes** |

<details> <summary>check_size Output</summary>

before:

```
File  .text    Size          Crate Name
2.1%  62.1%  8.6KiB      [Unknown] main
0.2%   5.4%    764B        gdbstub gdbstub::stub::state_machine::GdbStubStateMachineInner<gdbstub::stub::state_machine:...
0.1%   2.4%    336B        gdbstub gdbstub::protocol::commands::breakpoint::BasicBreakpoint::from_slice
0.1%   2.3%    320B        gdbstub <gdbstub::protocol::common::thread_id::ThreadId as core::convert::TryFrom<&[u8]>>::t...
0.1%   2.2%    308B        gdbstub gdbstub::protocol::common::hex::decode_hex_buf
0.1%   2.0%    284B        gdbstub gdbstub::protocol::response_writer::ResponseWriter<C>::write
0.1%   1.7%    236B        gdbstub gdbstub::stub::core_impl::resume::<impl gdbstub::stub::core_impl::GdbStubImpl<T,C>>:...
0.1%   1.6%    232B        gdbstub gdbstub::protocol::response_writer::ResponseWriter<C>::write_specific_thread_id
0.1%   1.5%    216B           core core::iter::traits::iterator::Iterator::nth
0.1%   1.5%    216B           core core::iter::traits::iterator::Iterator::nth
0.0%   1.1%    160B        gdbstub gdbstub::protocol::common::hex::decode_hex
0.0%   1.1%    156B        gdbstub gdbstub::protocol::common::hex::decode_hex
0.0%   1.1%    156B        gdbstub gdbstub::protocol::common::hex::decode_hex
0.0%   1.1%    156B        gdbstub gdbstub::protocol::response_writer::ResponseWriter<C>::write_num
0.0%   1.1%    152B        gdbstub gdbstub::protocol::response_writer::ResponseWriter<C>::inner_write
0.0%   1.0%    144B        gdbstub gdbstub::protocol::response_writer::ResponseWriter<C>::write_num
0.0%   1.0%    140B        gdbstub <gdbstub::protocol::common::thread_id::IdKind as core::convert::TryFrom<&[u8]>>::try...
0.0%   0.9%    132B        gdbstub gdbstub::protocol::response_writer::ResponseWriter<C>::flush
0.0%   0.9%    128B        gdbstub gdbstub::protocol::response_writer::ResponseWriter<C>::write_hex
0.0%   0.8%    116B gdbstub_nostd? <gdbstub_nostd::gdb::DummyTarget as gdbstub::target::ext::base::multithread::MultiTh...
0.0%   0.8%    108B           core <core::iter::adapters::skip::Skip<I> as core::iter::traits::iterator::Iterator>::next
0.0%   0.7%    104B           core <core::slice::iter::SplitMut<T,P> as core::iter::traits::iterator::Iterator>::next
0.0%   0.5%     76B gdbstub_nostd? <gdbstub_nostd::gdb::DummyTarget as gdbstub::target::ext::base::multithread::MultiTh...
0.0%   0.5%     76B gdbstub_nostd? <gdbstub_nostd::gdb::DummyTarget as gdbstub::target::ext::base::multithread::MultiTh...
0.0%   0.5%     76B gdbstub_nostd? <gdbstub_nostd::gdb::DummyTarget as gdbstub::target::ext::base::multithread::MultiTh...
0.0%   0.5%     68B   gdbstub_arch <gdbstub_arch::arm::reg::arm_core::ArmCoreRegs as gdbstub::arch::Registers>::gdb_des...
0.0%   0.4%     52B gdbstub_nostd? <gdbstub_nostd::gdb::DummyTarget as gdbstub::target::ext::base::multithread::MultiTh...
0.0%   0.4%     52B gdbstub_nostd? <gdbstub_nostd::gdb::DummyTarget as gdbstub::target::ext::base::multithread::MultiTh...
0.0%   0.4%     52B gdbstub_nostd? <gdbstub_nostd::gdb::DummyTarget as gdbstub::target::ext::base::multithread::MultiTh...
0.0%   0.4%     52B      [Unknown] _start
0.0%   0.3%     48B  gdbstub_nostd gdbstub_nostd::print_str::print_str
0.0%   0.1%     20B      [Unknown] call_weak_fn
0.0%   0.1%     12B gdbstub_nostd? <gdbstub_nostd::gdb::DummyTarget as gdbstub::target::ext::breakpoints::SwBreakpoint>...
3.3% 100.0% 13.8KiB                .text section size, the file size is 413.5KiB
section               size     addr
.interp                 27      568
.note.gnu.build-id      36      596
.note.ABI-tag           32      632
.gnu.hash               28      664
.dynsym                432      696
.dynstr                198     1128
.gnu.version            36     1326
.gnu.version_r          48     1368
.rela.dyn              192     1416
.rela.plt              312     1608
.init                   24     1920
.plt                   240     1952
.text                14128     2240
.fini                   20    16368
.rodata                670    16388
.eh_frame_hdr          300    17060
.eh_frame             1072    17360
.init_array              8   130384
.fini_array              8   130392
.dynamic               496   130400
.got                   176   130896
.data                    8   131072
.bss                     8   131080
.comment                87        0
Total                18586
```

after:

```
File  .text    Size          Crate Name
2.1%  62.2%  8.6KiB      [Unknown] main
0.2%   5.4%    764B        gdbstub gdbstub::stub::state_machine::GdbStubStateMachineInner<gdbstub::stub::state_machine:...
0.1%   2.4%    336B        gdbstub gdbstub::protocol::commands::breakpoint::BasicBreakpoint::from_slice
0.1%   2.3%    320B        gdbstub <gdbstub::protocol::common::thread_id::ThreadId as core::convert::TryFrom<&[u8]>>::t...
0.1%   2.2%    308B        gdbstub gdbstub::protocol::common::hex::decode_hex_buf
0.1%   2.0%    284B        gdbstub gdbstub::protocol::response_writer::ResponseWriter<C>::write
0.1%   1.7%    236B        gdbstub gdbstub::stub::core_impl::resume::<impl gdbstub::stub::core_impl::GdbStubImpl<T,C>>:...
0.1%   1.6%    232B        gdbstub gdbstub::protocol::response_writer::ResponseWriter<C>::write_specific_thread_id
0.1%   1.5%    216B           core core::iter::traits::iterator::Iterator::nth
0.1%   1.5%    216B           core core::iter::traits::iterator::Iterator::nth
0.0%   1.1%    160B        gdbstub gdbstub::protocol::common::hex::decode_hex
0.0%   1.1%    156B        gdbstub gdbstub::protocol::common::hex::decode_hex
0.0%   1.1%    156B        gdbstub gdbstub::protocol::common::hex::decode_hex
0.0%   1.1%    156B        gdbstub gdbstub::protocol::response_writer::ResponseWriter<C>::write_num
0.0%   1.1%    152B        gdbstub gdbstub::protocol::response_writer::ResponseWriter<C>::inner_write
0.0%   1.0%    144B        gdbstub gdbstub::protocol::response_writer::ResponseWriter<C>::write_num
0.0%   1.0%    140B        gdbstub <gdbstub::protocol::common::thread_id::IdKind as core::convert::TryFrom<&[u8]>>::try...
0.0%   0.9%    132B        gdbstub gdbstub::protocol::response_writer::ResponseWriter<C>::flush
0.0%   0.9%    128B        gdbstub gdbstub::protocol::response_writer::ResponseWriter<C>::write_hex
0.0%   0.8%    116B gdbstub_nostd? <gdbstub_nostd::gdb::DummyTarget as gdbstub::target::ext::base::multithread::MultiTh...
0.0%   0.8%    108B           core <core::iter::adapters::skip::Skip<I> as core::iter::traits::iterator::Iterator>::next
0.0%   0.7%    104B           core <core::slice::iter::SplitMut<T,P> as core::iter::traits::iterator::Iterator>::next
0.0%   0.5%     76B gdbstub_nostd? <gdbstub_nostd::gdb::DummyTarget as gdbstub::target::ext::base::multithread::MultiTh...
0.0%   0.5%     76B gdbstub_nostd? <gdbstub_nostd::gdb::DummyTarget as gdbstub::target::ext::base::multithread::MultiTh...
0.0%   0.5%     76B gdbstub_nostd? <gdbstub_nostd::gdb::DummyTarget as gdbstub::target::ext::base::multithread::MultiTh...
0.0%   0.5%     68B   gdbstub_arch <gdbstub_arch::arm::reg::arm_core::ArmCoreRegs as gdbstub::arch::Registers>::gdb_des...
0.0%   0.4%     52B gdbstub_nostd? <gdbstub_nostd::gdb::DummyTarget as gdbstub::target::ext::base::multithread::MultiTh...
0.0%   0.4%     52B gdbstub_nostd? <gdbstub_nostd::gdb::DummyTarget as gdbstub::target::ext::base::multithread::MultiTh...
0.0%   0.4%     52B gdbstub_nostd? <gdbstub_nostd::gdb::DummyTarget as gdbstub::target::ext::base::multithread::MultiTh...
0.0%   0.4%     52B      [Unknown] _start
0.0%   0.3%     48B  gdbstub_nostd gdbstub_nostd::print_str::print_str
0.0%   0.1%     20B      [Unknown] call_weak_fn
0.0%   0.1%     12B gdbstub_nostd? <gdbstub_nostd::gdb::DummyTarget as gdbstub::target::ext::breakpoints::SwBreakpoint>...
3.3% 100.0% 13.8KiB                .text section size, the file size is 414.2KiB
section               size     addr
.interp                 27      568
.note.gnu.build-id      36      596
.note.ABI-tag           32      632
.gnu.hash               28      664
.dynsym                432      696
.dynstr                198     1128
.gnu.version            36     1326
.gnu.version_r          48     1368
.rela.dyn              192     1416
.rela.plt              312     1608
.init                   24     1920
.plt                   240     1952
.text                14156     2240
.fini                   20    16396
.rodata                670    16416
.eh_frame_hdr          300    17088
.eh_frame             1072    17392
.init_array              8   130384
.fini_array              8   130392
.dynamic               496   130400
.got                   176   130896
.data                    8   131072
.bss                     8   131080
.comment                87        0
Total                18614
```

</details>

### DIFF
--- a/examples/armv4t_multicore/emu.rs
+++ b/examples/armv4t_multicore/emu.rs
@@ -36,9 +36,11 @@ pub enum Event {
     WatchRead(u32),
 }
 
+#[derive(PartialEq)]
 pub enum ExecMode {
     Step,
     Continue,
+    Stop,
 }
 
 /// incredibly barebones armv4t-based emulator
@@ -187,6 +189,10 @@ impl Emu {
         let mut evt = None;
 
         for id in [CpuId::Cpu, CpuId::Cop].iter().copied() {
+            if matches!(self.exec_mode.get(&id), Some(ExecMode::Stop)) {
+                continue;
+            }
+
             if let Some(event) = self.step_core(id) {
                 if evt.is_none() {
                     evt = Some((event, id));
@@ -198,26 +204,25 @@ impl Emu {
     }
 
     pub fn run(&mut self, mut poll_incoming_data: impl FnMut() -> bool) -> RunEvent {
-        // the underlying armv4t_multicore emulator runs both cores in lock step, so
-        // when GDB requests a specific core to single-step, all we need to do is jot
-        // down that we want to single-step the system, as there is no way to
-        // single-step a single core while the other runs.
+        // The underlying armv4t_multicore emulator cycles all cores in lock-step.
         //
-        // In more complex emulators / implementations, this simplification is _not_
-        // valid, and you should track which specific TID the GDB client requested to be
-        // single-stepped, and run them appropriately.
+        // Inside `self.step()`, we iterate through all cores and only invoke
+        // `step_core` if that core's `ExecMode` is not `Stop`.
 
-        let should_single_step = matches!(
-            self.exec_mode
-                .get(&CpuId::Cpu)
-                .or_else(|| self.exec_mode.get(&CpuId::Cop)),
-            Some(&ExecMode::Step)
-        );
+        let should_single_step = self.exec_mode.values().any(|mode| mode == &ExecMode::Step);
 
         match should_single_step {
             true => match self.step() {
                 Some((event, id)) => RunEvent::Event(event, id),
-                None => RunEvent::Event(Event::DoneStep, CpuId::Cpu),
+                None => {
+                    let stepping_core = self
+                        .exec_mode
+                        .iter()
+                        .find(|&(_, mode)| mode == &ExecMode::Step)
+                        .map(|(id, _)| *id)
+                        .unwrap_or(CpuId::Cpu);
+                    RunEvent::Event(Event::DoneStep, stepping_core)
+                }
             },
             false => {
                 let mut cycles = 0;
--- a/src/stub/core_impl/resume.rs
+++ b/src/stub/core_impl/resume.rs
@@ -166,6 +166,13 @@ impl<T: Target, C: Connection> GdbStubImpl<T, C> {
     ) -> Result<(), Error<T::Error, C::Error>> {
         ops.clear_resume_actions().map_err(Error::TargetError)?;
 
+        // Track whether the packet contains a wildcard/default continue action
+        // (e.g., `c` or `c:-1`).
+        //
+        // Presence of this action implies "Scheduler Locking" is OFF.
+        // Absence implies "Scheduler Locking" is ON.
+        let mut has_wildcard_continue = false;
+
         for action in actions.iter() {
             use crate::protocol::commands::_vCont::VContKind;
 
@@ -185,6 +192,7 @@ impl<T: Target, C: Connection> GdbStubImpl<T, C> {
                         None | Some(SpecificIdKind::All) => {
                             // Target API contract specifies that the default
                             // resume action for all threads is continue.
+                            has_wildcard_continue = true;
                         }
                         Some(SpecificIdKind::WithId(tid)) => ops
                             .set_resume_action_continue(tid, signal)
@@ -249,6 +257,15 @@ impl<T: Target, C: Connection> GdbStubImpl<T, C> {
                     return Err(Error::PacketUnexpected);
                 }
             }
+        }
+
+        if !has_wildcard_continue {
+            let Some(locking_ops) = ops.support_scheduler_locking() else {
+                return Err(Error::MissingMultiThreadSchedulerLocking);
+            };
+            locking_ops
+                .set_resume_action_scheduler_lock()
+                .map_err(Error::TargetError)?;
         }
 
         ops.resume().map_err(Error::TargetError)
--- a/src/target/ext/base/multithread.rs
+++ b/src/target/ext/base/multithread.rs
@@ -134,9 +134,21 @@ pub trait MultiThreadResume: Target {
     /// GDB client had specified using any of the `set_resume_action_XXX`
     /// methods.
     ///
-    /// Any thread that wasn't explicitly resumed by a `set_resume_action_XXX`
-    /// method should be resumed as though it was resumed with
-    /// `set_resume_action_continue`.
+    /// # Default Resume Behavior
+    ///
+    /// By default, any thread that wasn't explicitly resumed by a
+    /// `set_resume_action_XXX` method should be resumed as though it was
+    /// resumed with `set_resume_action_continue`.
+    ///
+    /// **However**, if [`support_scheduler_locking`] is implemented and
+    /// [`set_resume_action_scheduler_lock`] has been called for the current
+    /// resume cycle, this default changes: **unmentioned threads must remain
+    /// stopped.**
+    ///
+    /// [`support_scheduler_locking`]: Self::support_scheduler_locking
+    /// [`set_resume_action_scheduler_lock`]: MultiThreadSchedulerLocking::set_resume_action_scheduler_lock
+    ///
+    /// # Protocol Extensions
     ///
     /// A basic target implementation only needs to implement support for
     /// `set_resume_action_continue`, with all other resume actions requiring
@@ -146,14 +158,17 @@ pub trait MultiThreadResume: Target {
     /// ----------------------------|------------------------------
     /// Optimized [Single Stepping] | See [`support_single_step()`]
     /// Optimized [Range Stepping]  | See [`support_range_step()`]
+    /// [Scheduler Locking]         | See [`support_scheduler_locking()`]
     /// "Stop"                      | Used in "Non-Stop" mode \*
     ///
     /// \* "Non-Stop" mode is currently unimplemented in `gdbstub`
     ///
     /// [Single stepping]: https://sourceware.org/gdb/current/onlinedocs/gdb/Continuing-and-Stepping.html#index-stepi
     /// [Range Stepping]: https://sourceware.org/gdb/current/onlinedocs/gdb/Continuing-and-Stepping.html#range-stepping
+    /// [Scheduler Locking]: https://sourceware.org/gdb/current/onlinedocs/gdb#index-scheduler-locking-mode
     /// [`support_single_step()`]: Self::support_single_step
     /// [`support_range_step()`]: Self::support_range_step
+    /// [`support_scheduler_locking()`]: Self::support_scheduler_locking
     ///
     /// # Additional Considerations
     ///
@@ -233,6 +248,14 @@ pub trait MultiThreadResume: Target {
     ) -> Option<super::reverse_exec::ReverseContOps<'_, Tid, Self>> {
         None
     }
+
+    /// Support for [scheduler locking].
+    ///
+    /// [scheduler locking]: https://sourceware.org/gdb/current/onlinedocs/gdb#index-scheduler-locking-mode
+    #[inline(always)]
+    fn support_scheduler_locking(&mut self) -> Option<MultiThreadSchedulerLockingOps<'_, Self>> {
+        None
+    }
 }
 
 define_ext!(MultiThreadResumeOps, MultiThreadResume);
@@ -290,3 +313,26 @@ pub trait MultiThreadRangeStepping: Target + MultiThreadResume {
 }
 
 define_ext!(MultiThreadRangeSteppingOps, MultiThreadRangeStepping);
+
+/// Target Extension - support for GDB's "Scheduler Locking" mode.
+/// See [`MultiThreadResume::support_scheduler_locking`].
+pub trait MultiThreadSchedulerLocking: Target + MultiThreadResume {
+    /// Configure the target to enable "Scheduler Locking" for the upcoming
+    /// resume.
+    ///
+    /// This method is invoked when the GDB client expects only a specific set
+    /// of threads to run, while all other threads remain frozen. This behavior
+    /// is typically toggled in GDB using the `set scheduler-locking on`
+    /// command.
+    ///
+    /// When this method is called, the implementation must ensure that any
+    /// threads not explicitly resumed via previous `set_resume_action_...`
+    /// calls **remain stopped**.
+    ///
+    /// This prevents any "implicit continue" behavior for unmentioned threads,
+    /// satisfying GDB's expectation that only the designated threads will
+    /// advance during the next resume.
+    fn set_resume_action_scheduler_lock(&mut self) -> Result<(), Self::Error>;
+}
+
+define_ext!(MultiThreadSchedulerLockingOps, MultiThreadSchedulerLocking);
--- a/examples/armv4t_multicore/gdb.rs
+++ b/examples/armv4t_multicore/gdb.rs
@@ -185,6 +185,13 @@ impl MultiThreadResume for Emu {
 
         Ok(())
     }
+
+    #[inline(always)]
+    fn support_scheduler_locking(
+        &mut self,
+    ) -> Option<target::ext::base::multithread::MultiThreadSchedulerLockingOps<'_, Self>> {
+        Some(self)
+    }
 }
 
 impl target::ext::base::multithread::MultiThreadSingleStep for Emu {
@@ -291,6 +298,15 @@ impl target::ext::thread_extra_info::ThreadExtraInfo for Emu {
         let info = format!("CPU {:?}", cpu_id);
 
         Ok(copy_to_buf(info.as_bytes(), buf))
+    }
+}
+
+impl target::ext::base::multithread::MultiThreadSchedulerLocking for Emu {
+    fn set_resume_action_scheduler_lock(&mut self) -> Result<(), Self::Error> {
+        for id in [CpuId::Cpu, CpuId::Cop] {
+            self.exec_mode.entry(id).or_insert(ExecMode::Stop);
+        }
+        Ok(())
     }
 }
 
--- a/src/stub/error.rs
+++ b/src/stub/error.rs
@@ -43,6 +43,7 @@ pub(crate) enum InternalError<T, C> {
     MissingCurrentActivePidImpl,
     TracepointFeatureUnimplemented(u8),
     TracepointUnsupportedSourceEnumeration,
+    MissingMultiThreadSchedulerLocking,
 
     // Internal - A non-fatal error occurred (with errno-style error code)
     //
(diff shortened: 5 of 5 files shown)


examples/armv4t_multicore/emu.rs
```diff
@@ -207,17 +213,20 @@ impl Emu {
         // valid, and you should track which specific TID the GDB client requested to be
         // single-stepped, and run them appropriately.
```
daniel5151: would it be safe to remove this old code comment now that you've added the ability to only step a single core at a time?

src/target/ext/base/multithread.rs
```diff
@@ -290,3 +313,25 @@ pub trait MultiThreadRangeStepping: Target + MultiThreadResume {
...
 
 define_ext!(MultiThreadRangeSteppingOps, MultiThreadRangeStepping);
+
+/// Target Extension - support for GDB's "Scheduler Locking" mode.
+/// See [`MultiThreadResume::support_scheduler_locking`].
+pub trait MultiThreadSchedulerLocking: Target + MultiThreadResume {
+    /// Configure the target to enable "Scheduler Locking" for the upcoming
+    /// resume.
+    ///
+    /// This method is called when the GDB client sends a `vCont` packet that
```
daniel5151: In general, `gdbstub`'s docs try to avoid leaking the nitty-gritty of the underlying protocol to the end user. Interested users can always crack the code open to figure out these details... but these docs are expected to answer high-level "what is this, how is this useful" questions.

Instead - I would remove this paragraph, and maybe include some phrasing that mentions that this behavior is toggled on/off in GDB by using the `set scheduler-locking {on/off}` command.

src/target/ext/base/multithread.rs
```diff
@@ -290,3 +313,25 @@ pub trait MultiThreadRangeStepping: Target + MultiThreadResume {
...
+    /// **does not** include a default "continue" action (e.g.,
+    /// `vCont;s:pPID.TID` without a trailing `;c`).
+    ///
+    /// When this method is called, the implementation **must ensure that any
+    /// threads not explicitly resumed** via previous
+    /// `set_resume_action_...` calls **remain stopped**.
```
daniel5151:
```suggestion
    /// When this method is called, the implementation must ensure that any
    /// threads not explicitly resumed via previous
    /// `set_resume_action_...` calls **remain stopped**.
```

no need to double emphasis here

daniel5151 (changes requested):
Great work! Overall design looks good, so after we fix up a few minor doc and error-handling related things, we should be gtg here.

Also, like I mentioned in https://github.com/daniel5151/gdbstub/issues/178#issuecomment-3683126118, it'd be good to double-check that the approach we're talking here lines up with how the GDB client (and maybe `gdbserver`) handle `scheduler-locking`.

I'll see if I can find some time to poke around https://github.com/bminor/binutils-gdb myself, but would def appreciate it if you could poke around a bit as part of this diff.

* * *

Best of luck with exams! I'm also somewhat busy this week (home for the holidays and whatnot), so no need to rush or anything.

daniel5151:
> Also, like I mentioned in [#178 (comment)](https://github.com/daniel5151/gdbstub/issues/178#issuecomment-3683126118), it'd be good to double-check that the approach we're talking here lines up with how the GDB client (and maybe `gdbserver`) handle `scheduler-locking`.

I nerd sniped myself, so I did some digging here lol.

https://github.com/bminor/binutils-gdb/blob/b0196f0/gdb/remote.c#L960-L961

It _looks_ like GDB unconditionally assumes that the remote target supports scheduler-locking, which seems... not ideal. Interestingly enough - I chased the `git blame` chain to see when this default was introduced, and it looks like it stretches waaaaaaaaaay back into the past (I'm talking before 1999, and even before the initial commit that migrated GDB over to git). Check it out: https://github.com/bminor/binutils-gdb/blob/c906108c21474dfb4ed285bcc0ac6fe02cd400cc/gdb/remote.c#L3389

So, if you want some bonus points - you might want to consider opening an issue (or even sending in a PR?) to GDB that points out this behavior, and that it might be more user-friendly if GDB negotiated support for scheduler-locking on remote targets via a `qSupported` feature.

src/stub/core_impl/resume.rs
```diff
@@ -251,6 +259,18 @@ impl<T: Target, C: Connection> GdbStubImpl<T, C> {
             }
         }
 
+        if !has_wildcard_continue {
+            let Some(locking_ops) = ops.support_scheduler_locking() else {
+                error!(
+                    "GDB requested Scheduler Locking (vCont packet missing wildcard action), but the Target does not implement the `MultiThreadSchedulerLocking` IDET."
+                );
+                return Err(Error::PacketUnexpected);
```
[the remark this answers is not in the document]
Satar07: Fixed! Added the InternalError variant. Thanks for the tip, learned a lot about the error handling here

daniel5151 (approved):

marc-hb:
> It looks like GDB unconditionally assumes that the remote target supports scheduler-locking, which seems... not ideal.

Understatement?
- #200

PS: I did not read the whole thread, apologies. I only spotted this line which I found... interesting :-)

marc-hb:
> First off, thanks for the deep dive! Reading that massive amount of ancient C code is honestly painful to look at 😂.

... and it got even more painful now that https://github.com/bminor/binutils-gdb/ apparently vanished! Any good alternative to recommend? One where "translating" the above URLs isn't too painful.

EDIT, forgot: the "Non-Stop multi-threaded debugging in GDB" paper by Nathan Sidwell et al. seems significantly easier to read than source code.

> Since the Remote Serial Protocol uses wildcards in its data packets to convey information, there seems to be an implicit assumption that the remote server should possess the capability to control threads/processes to that degree. In that sense, the hardcoded behavior (defaulting to assuming support) actually seems reasonable
> [...]
> Indeed, I also saw that this was a purely client-side concept.

If this is indeed a purely client-side concept without any client-server negociations, then could either of you speculate why `set scheduler-locking on` is NOT an effective workaround for https://github.com/probe-rs/probe-rs/issues/3805?

While slowly digesting all this context I found the following code in this PR:

```rust
if !has_wildcard_continue {
            let Some(locking_ops) = ops.support_scheduler_locking() else {
                return Err(Error::MissingMultiThreadSchedulerLocking);
            };
```

Naively, optimistically and... wrongly(?), `set scheduler-locking on` should NOT send a continue wildcard; that's basically its job. So, the above code should not error anymore. Yet it still does.

Wait: `set scheduler-locking on` fails even faster than `off`! (the default) The `Error::MissingMultiThreadSchedulerLocking` error gets throw on the very first `continue` while it gets thrown on the _second_ `continue` after one successful breakpoint when `set scheduler-locking off`. What is going on?

The first `continue` that either fails or works is: releasing from a default `monitor reset halt` configured with the following in `Embed.toml`:
```ini
[default.reset]
halt_afterwards = true
```

Any simple way to trace the GDB packets? That should be easier than reading the GDB code. Or... debugging GDB itself! Wireshark maybe? Some verbose `gdbstub` logging?

Sorry to insist but it's really frustrating not to be able to avoid a _multithread_ problem on a microcontroller running a single "Hello world" thread. The "purely client-side concept" observation above just rubbed even more salt on that wound :-)

marc-hb:
I don't know whether this is relevant, but sharing just in case:

```
Reading symbols from ../../../target/thumbv7em-none-eabihf/debug/examples/init...
(gdb) set non-stop on
(gdb) target remote :1337
Remote debugging using :1337
❌️ Non-stop mode requested, but remote does not support non-stop
(gdb) 
```
On the gdbstub side:
```
     GDB stub listening at 127.0.0.1:1337

  < waits until target remote connects>

During the execution of GDB an error was encountered:
failed to fill whole buffer
        Done processing config profile default
```

Satar07:
The `MultiThreadSchedulerLocking` trait is specifically designed to handle the `scheduler-locking on` scenario. The frustrating part is that GDB doesn't actually negotiate the scheduler state with the server. It just sends `continue` or `step` commands for either specific threads or all threads (using wildcards). As the server, we only see these packets—we don't get a flag saying "scheduler locking is now on".

Because of this, I made a functional assumption in the implementation: if a `continue` packet arrives *without* a wildcard, it means GDB expects scheduler locking. In that moment, the server absolutely needs fine-grained control to keep the other threads suspended. If `scheduler-locking` is on, but the server hasn't implemented this trait, it *has* to throw an error by design. Otherwise, `gdbstub`'s default behavior would just resume all threads anyway, completely breaking the user's lock request.

This explains why `set scheduler-locking on` fails on the very first `continue`. Once you hit that initial breakpoint (or reset halt), GDB has a specific thread selected. With locking turned on, it sends a `continue` strictly for that thread, omitting the wildcard. `gdbstub` sees no wildcard, checks for the trait, doesn't find it, and immediately throws the `MissingMultiThreadSchedulerLocking` error.

The behavior you're seeing with `scheduler-locking off`—where it works once and then fails on the *second* continue—is honestly outside what I originally anticipated. Normally, with it off, GDB should keep appending that wildcard, so the trait shouldn't be triggered at all.

I can only guess why this happens. It's possible that for embedded targets, once GDB hits a breakpoint or gathers enough info to realize there's only a single thread/process, it stops sending the wildcard for subsequent commands and implicitly acts like it's locked. I'm not 100% sure about this, though. Grabbing the packet logs by running with `RUST_LOG=trace` (or `debug`) would definitely clear up what GDB is actually sending.

Part of the underlying issue here is that `probe-rs` is using `gdbstub`'s multi-thread implementation to handle single-threaded targets, which drags in all this multi-thread locking complexity where it isn't really needed.

Looking back, inferring a "lock" state from packet wildcards might just be a flawed design on our end. It would probably be cleaner to focus purely on raw thread start/stop signals rather than trying to map a conceptual lock state. We went with this approach as a compromise to keep the existing API stable for now. There's ongoing work to overhaul this in [#197](https://github.com/daniel5151/gdbstub/pull/197), which will hopefully put things in a much better spot.

As for a workaround/fix for your specific situation, I've outlined it over in [probe-rs/probe-rs#3805](https://github.com/probe-rs/probe-rs/issues/3805)

Issue #196, opened 2026-05-05
Remove / Rename synthetic `DoneStep` stop reason


jonathanzetier:
While developing a PR for multiprocess support and testing it with the `armv4t_multithreaded` example project, I found that every time I switched to thread 2 (the coprocessor thread) and stepped, the client would switch back to thread 1 (the cpu thread).  Re-checking out `dev/0.8` and I confirmed it still exists.

With `set debug remote on` in the GDB client, here's the most relevant excerpt that shows stepping thread 2 (while continuing the other processes):

```
[remote] Sending packet: $vCont;s:p1.2;c:p1.-1#f8
[remote] wait: enter
[remote] wait: exit
[remote] wait: enter
  [remote] Packet received: S05
  [remote] select_thread_for_ambiguous_stop_reply: enter
    [remote] select_thread_for_ambiguous_stop_reply: process_wide_stop = 0
    [remote] select_thread_for_ambiguous_stop_reply: first resumed thread is Thread 1.1
    [remote] select_thread_for_ambiguous_stop_reply: is this guess ambiguous? = 1
warning: multi-threaded target stopped without sending a thread-id, using first non-exited thread
  [remote] select_thread_for_ambiguous_stop_reply: exit
```

Granted, this was sent with `scheduler-locking mode off` (with `scheduler-locking mode on`, thread 2 remains the active thread), which means that the scheduler *could* schedule anything, and so I'm not complaining about the fact that the thread switches in this case, for `armv4t_multithreaded`.

However, stubs might be able to provide more information to the GDB client (and this also happens when I step with `gdb` and `gdbserver` locally): the stub might be able to say that the step concluded and also say that the thread that was requested to be stepped actually was.  This should make the GDB client user's experience a little better, since the likeliest case of a step is that the step concluded before any other thread could stop, and the user wouldn't have to keep switching back to the thread they're actually trying to work with.

It seems like giving `DoneStep` an `Option<Tid>` (or adding a `DoneStepWithThread`, like with `Signal`) would give stubs the ability to do this.  If we're already breaking the API in v0.8 for multiprocess support, this might be another quick+simple to migrate breakage we could sneak in too.  A current workaround exists of issuing a `SwBreak` or a `HwBreak` instead of a `DoneStep`, but this isn't intuitive and seems like it would be easier for `Target` implementations to misuse.

daniel5151:
Truthfully, I think a better approach would be to remove the DoneStep stop reason entirely 😅 

[...]

Issue #124, opened 2023-02-28
Multiprocess debugging


daniel5151:
`gdbstub` already implements protocol-level multiprocess extensions "under-the-hood", and simply hard-codes a fake PID in single/multi-threaded mode (much the same way it hard-codes a fake TID in single-threaded mode).

Adding true multiprocess debugging will most likely involve adding a new `target::ext::base::multiprocess` API set, along with doing the requisite plumbing to track and report the current PID.

In addition, we'd want to add a new in-tree `armv4t_multiprocess` example, which wouldn't _actually_ spin up multiple processes, but would instead spin up multiple multi-core `armv4t` emulator instances, just so there's an in-tree example of this stuff working.

Why API-breaking? Unfortunately, I neglected to mark the `BaseOps` enum as `non_exhaustive`, which means implementing this will be a breaking change. Aside from that, this likely could've been entirely non-breaking...

xobs:
I think the only thing that needs to be implemented is `info os processes`:

```
(gdb) info os processes
[remote] Sending packet: $qXfer:osdata:read:processes:0,1000#fa
[remote] Packet received: l<osdata type="processes">\n<item><column name="pid">1</column><column name="user">root</column><column name="command">/init </column><column name="cores">6,15</column></item><item><column name="pid">4</column><column name="user">root</column><column name="command">plan9 --control-socket 5 --log-level 4 --server-fd 6 --pipe-fd 8 --log-truncate </column><column name="cores">5,19</column></item><item><column name="pid">7</column><column name="user">root</column><column name="command">/init </column><column nam [2029 bytes omitted]
pid        user       command    cores
1          root       /init      6,15
4          root       plan9 --control-socket 5 --log-level 4 --server-fd 6 --pipe-fd 8 --log-truncate 5,19
7          root       /init      16
8          root       /init      2
9          user       -bash      23
10         root       /init      11
11         root       /init      11
12         user       -bash      13
468        user       dbus-launch --autolaunch 857863f4cf286a4142beec6a62a03f1f --binary-syntax --close-stderr 20
469        user       /usr/bin/dbus-daemon --syslog-only --fork --print-pid 5 --print-address 7 --session 23
499        user       gdb        0,1,2,4,5,6,8,9,10,12,13,14,15,16,17,18,20,21,22,23
525        root       /init      17
526        root       /init      19
527        user       -bash      21
544        user       gdbserver --multi localhost:3454 14
(gdb)
```

[...]

PR #197, opened 2026-05-06
Rework scheduler locking

jonathanzetier:
### Description

This lays down some groundwork for v0.8, where the primary new feature will be multiprocess support, which will be similar to multithreaded support.

As discussed on the multiprocess github support issue, there's a simpler way to support the scheduler locking behavior that we plan to use for multiprocess targets, and since we're breaking the API anyways with v0.8, it's a decent time to implement it for multithreaded targets as well.

While I was in the area, I fixed some typos I noticed when I had copied most of these during local development of multiprocess-extensions.

This PR does NOT do anything about the proposed changes in this comment regarding the order in which we parse resume actions laid out here: https://github.com/daniel5151/gdbstub/issues/124#issuecomment-4374028370.

### API Stability

This is part of the initial transition to v0.8.  Justification for breaking the API is provided here: https://github.com/daniel5151/gdbstub/issues/124#issuecomment-4375276483

Another quick API breakage that might make sense to throw in this PR too is changing the name of `is_thread_alive` to `check_thread_alive` (available for bikeshedding), which removes the need for the `#[allow(clippy::wrong_self_convention)]`.

### Checklist

- Documentation
  - [N/A] (if appropriate) Added feature to "Debugging Features" in README.md
- Validation
  - [ ] Included output of running `examples/armv4t` with `RUST_LOG=trace` + any relevant GDB output under the "Validation" section below <see below>
  - [x] Included output of running `./example_no_std/check_size.sh` before/after changes under the "Validation" section below
  - [N/A] Included a basic sample implementation in `examples/armv4t`
  - [N/A] IDET can be optimized out (confirmed via `./example_no_std/check_size.sh`)
  - [N/A] **OR** implementation requires introducing non-optional binary bloat (please elaborate under "Description")
  - [N/A] I have tested this code in my project, and to the best of my knowledge, it is working as intended.

### Validation

I tested a lot of different scenarios in armv4t for both `dev/0.8` and this PR.  I'm taking a "ask forgiveness, not permission" here approach for the logs; there's a lot of logs to generate (which also means a lot of logs for you to sift through), so for now I'll list all the scenarios I tested, and I can post logs on request (or test alternative scenarios).

* scheduler-lock off, `continue` from thread 1
* scheduler-lock off, `continue` from thread 2
* scheduler-lock off, `nexti 10` from thread 1
* scheduler-lock off, `nexti 80` from thread 2 (which led to the discovery of #196)
* scheduler-lock on, `continue` from thread 1 (program hangs; I needed to ctrl+C, switch to thread 2, `continue`, ctrl+C, switch to thread 1, `continue`, which makes sense given the behavior of `scheduler-lock` and the example binary)
* scheduler-lock on, `continue` from thread 2 (program hangs; ctrl+C'ing and switching between threads doesn't actually seem to ever resolve.  When I have more time I'd like  dig into the example binary to figure out why that might be and if there's a deeper underlying bug, but this is the behavior exhibited by `dev/0.8`, so for the purposes of this PR I stopped investigating there)
* scheduler-lock on, `nexti 10` from thread 1
* scheduler-lock on, `nexti 80` from thread 2

Let me know if you would like logs for any (...or all) of these, or if there are any situations I didn't consider.

<details>
<summary>Before/After `./example_no_std/check_size.sh` output</summary>

### Before

```
File  .text    Size          Crate Name
2.5%  68.8% 10.4KiB      [Unknown] main
0.2%   6.3%    975B        gdbstub gdbstub::stub::state_machine::GdbStubStateMachineInner<gdbstub::stub::state_machine::state::Running,T,C>::report_stop
0.1%   2.4%    372B        gdbstub gdbstub::protocol::commands::breakpoint::BasicBreakpoint::from_slice
0.1%   1.9%    295B        gdbstub <gdbstub::protocol::common::thread_id::ThreadId as core::convert::TryFrom<&[u8]>>::try_from
0.1%   1.8%    275B        gdbstub gdbstub::protocol::common::hex::decode_hex_buf
0.1%   1.4%    222B        gdbstub gdbstub::stub::core_impl::resume::<impl gdbstub::stub::core_impl::GdbStubImpl<T,C>>::write_stop_common
0.1%   1.4%    221B        gdbstub gdbstub::protocol::response_writer::ResponseWriter<C>::write
0.0%   1.3%    204B        gdbstub gdbstub::protocol::response_writer::ResponseWriter<C>::write_specific_thread_id
0.0%   0.9%    143B           core core::iter::traits::iterator::Iterator::nth
0.0%   0.9%    143B           core core::iter::traits::iterator::Iterator::nth
0.0%   0.9%    132B        gdbstub gdbstub::protocol::common::hex::decode_hex
0.0%   0.8%    131B        gdbstub gdbstub::protocol::common::hex::decode_hex
0.0%   0.8%    122B        gdbstub gdbstub::protocol::response_writer::ResponseWriter<C>::inner_write
0.0%   0.7%    110B        gdbstub gdbstub::protocol::common::hex::decode_hex
0.0%   0.7%    109B        gdbstub gdbstub::protocol::response_writer::ResponseWriter<C>::write_num
0.0%   0.7%    107B        gdbstub gdbstub::protocol::response_writer::ResponseWriter<C>::write_num
0.0%   0.7%    106B gdbstub_nostd? <gdbstub_nostd::gdb::DummyTarget as gdbstub::target::ext::base::multithread::MultiThreadBase>::read_addrs
0.0%   0.7%    104B        gdbstub gdbstub::protocol::response_writer::ResponseWriter<C>::write_hex
0.0%   0.7%    102B        gdbstub gdbstub::protocol::response_writer::ResponseWriter<C>::flush
0.0%   0.6%     93B           core <core::iter::adapters::skip::Skip<I> as core::iter::traits::iterator::Iterator>::next
0.0%   0.6%     92B        gdbstub <gdbstub::protocol::common::thread_id::IdKind as core::convert::TryFrom<&[u8]>>::try_from
0.0%   0.6%     87B           core <core::slice::iter::SplitMut<T,P> as core::iter::traits::iterator::Iterator>::next
0.0%   0.4%     67B   gdbstub_arch <gdbstub_arch::arm::reg::arm_core::ArmCoreRegs as gdbstub::arch::Registers>::gdb_deserialize
0.0%   0.4%     65B gdbstub_nostd? <gdbstub_nostd::gdb::DummyTarget as gdbstub::target::ext::base::multithread::MultiThreadBase>::write_registers
0.0%   0.4%     65B gdbstub_nostd? <gdbstub_nostd::gdb::DummyTarget as gdbstub::target::ext::base::multithread::MultiThreadBase>::read_registers
0.0%   0.4%     65B gdbstub_nostd? <gdbstub_nostd::gdb::DummyTarget as gdbstub::target::ext::base::multithread::MultiThreadBase>::write_addrs
0.0%   0.3%     50B gdbstub_nostd? <gdbstub_nostd::gdb::DummyTarget as gdbstub::target::ext::base::multithread::MultiThreadResume>::resume
0.0%   0.3%     50B gdbstub_nostd? <gdbstub_nostd::gdb::DummyTarget as gdbstub::target::ext::base::multithread::MultiThreadResume>::set_resume_action_continue
0.0%   0.3%     50B gdbstub_nostd? <gdbstub_nostd::gdb::DummyTarget as gdbstub::target::ext::base::multithread::MultiThreadResume>::clear_resume_actions
0.0%   0.3%     44B  gdbstub_nostd gdbstub_nostd::print_str::print_str
0.0%   0.2%     38B      [Unknown] _start
0.0%   0.0%      6B gdbstub_nostd? <gdbstub_nostd::gdb::DummyTarget as gdbstub::target::ext::breakpoints::SwBreakpoint>::add_sw_breakpoint
3.6% 100.0% 15.1KiB                .text section size, the file size is 420.9KiB
section               size    addr
.interp                 28     680
.note.ABI-tag           32     708
.note.gnu.build-id      36     740
.dynsym                384     776
.gnu.version            32    1160
.gnu.version_r          64    1192
.gnu.hash               28    1256
.dynstr                215    1284
.rela.dyn              432    1504
.rela.plt               24    1936
.rodata               1005    1968
.eh_frame_hdr          268    2976
.eh_frame             1328    3248
.text                15495    8672
.init                   27   24168
.fini                   13   24196
.plt                    32   24224
.fini_array              8   28352
.init_array              8   28360
.dynamic               432   28368
.got                   120   28800
.got.plt                32   28920
.relro_padding        3816   28952
.tm_clone_table          0   33048
.data                    8   33048
.bss                     1   33056
.comment               184       0
Total                24052
```

### After

```
File  .text    Size          Crate Name
2.5%  68.5% 10.3KiB      [Unknown] main
0.2%   6.3%    975B        gdbstub gdbstub::stub::state_machine::GdbStubStateMachineInner<gdbstub::stub::state_machine::state::Running,T,C>::report_stop
0.1%   2.4%    372B        gdbstub gdbstub::protocol::commands::breakpoint::BasicBreakpoint::from_slice
0.1%   1.9%    295B        gdbstub <gdbstub::protocol::common::thread_id::ThreadId as core::convert::TryFrom<&[u8]>>::try_from
0.1%   1.8%    275B        gdbstub gdbstub::protocol::common::hex::decode_hex_buf
0.1%   1.4%    222B        gdbstub gdbstub::stub::core_impl::resume::<impl gdbstub::stub::core_impl::GdbStubImpl<T,C>>::write_stop_common
0.1%   1.4%    221B        gdbstub gdbstub::protocol::response_writer::ResponseWriter<C>::write
0.0%   1.3%    204B        gdbstub gdbstub::protocol::response_writer::ResponseWriter<C>::write_specific_thread_id
0.0%   0.9%    143B           core core::iter::traits::iterator::Iterator::nth
0.0%   0.9%    143B           core core::iter::traits::iterator::Iterator::nth
0.0%   0.9%    132B        gdbstub gdbstub::protocol::common::hex::decode_hex
0.0%   0.9%    131B        gdbstub gdbstub::protocol::common::hex::decode_hex
0.0%   0.8%    122B        gdbstub gdbstub::protocol::response_writer::ResponseWriter<C>::inner_write
0.0%   0.7%    110B        gdbstub gdbstub::protocol::common::hex::decode_hex
0.0%   0.7%    109B        gdbstub gdbstub::protocol::response_writer::ResponseWriter<C>::write_num
0.0%   0.7%    107B        gdbstub gdbstub::protocol::response_writer::ResponseWriter<C>::write_num
0.0%   0.7%    106B gdbstub_nostd? <gdbstub_nostd::gdb::DummyTarget as gdbstub::target::ext::base::multithread::MultiThreadBase>::read_addrs
0.0%   0.7%    104B        gdbstub gdbstub::protocol::response_writer::ResponseWriter<C>::write_hex
0.0%   0.7%    102B        gdbstub gdbstub::protocol::response_writer::ResponseWriter<C>::flush
0.0%   0.6%     93B           core <core::iter::adapters::skip::Skip<I> as core::iter::traits::iterator::Iterator>::next
0.0%   0.6%     92B        gdbstub <gdbstub::protocol::common::thread_id::IdKind as core::convert::TryFrom<&[u8]>>::try_from
0.0%   0.6%     87B           core <core::slice::iter::SplitMut<T,P> as core::iter::traits::iterator::Iterator>::next
0.0%   0.4%     67B   gdbstub_arch <gdbstub_arch::arm::reg::arm_core::ArmCoreRegs as gdbstub::arch::Registers>::gdb_deserialize
0.0%   0.4%     65B gdbstub_nostd? <gdbstub_nostd::gdb::DummyTarget as gdbstub::target::ext::base::multithread::MultiThreadBase>::write_registers
0.0%   0.4%     65B gdbstub_nostd? <gdbstub_nostd::gdb::DummyTarget as gdbstub::target::ext::base::multithread::MultiThreadBase>::read_registers
0.0%   0.4%     65B gdbstub_nostd? <gdbstub_nostd::gdb::DummyTarget as gdbstub::target::ext::base::multithread::MultiThreadBase>::write_addrs
0.0%   0.3%     50B gdbstub_nostd? <gdbstub_nostd::gdb::DummyTarget as gdbstub::target::ext::base::multithread::MultiThreadResume>::resume
0.0%   0.3%     50B gdbstub_nostd? <gdbstub_nostd::gdb::DummyTarget as gdbstub::target::ext::base::multithread::MultiThreadResume>::set_resume_action_continue
0.0%   0.3%     50B gdbstub_nostd? <gdbstub_nostd::gdb::DummyTarget as gdbstub::target::ext::base::multithread::MultiThreadResume>::clear_resume_actions
0.0%   0.3%     44B  gdbstub_nostd gdbstub_nostd::print_str::print_str
0.0%   0.2%     38B      [Unknown] _start
0.0%   0.0%      6B gdbstub_nostd? <gdbstub_nostd::gdb::DummyTarget as gdbstub::target::ext::breakpoints::SwBreakpoint>::add_sw_breakpoint
3.6% 100.0% 15.0KiB                .text section size, the file size is 418.3KiB
section               size    addr
.interp                 28     680
.note.ABI-tag           32     708
.note.gnu.build-id      36     740
.dynsym                384     776
.gnu.version            32    1160
.gnu.version_r          64    1192
.gnu.hash               28    1256
.dynstr                215    1284
.rela.dyn              432    1504
.rela.plt               24    1936
.rodata                989    1968
.eh_frame_hdr          268    2960
.eh_frame             1328    3232
.text                15357    8656
.init                   27   24016
.fini                   13   24044
.plt                    32   24064
.fini_array              8   28192
.init_array              8   28200
.dynamic               432   28208
.got                   120   28640
.got.plt                32   28760
.relro_padding        3976   28792
.tm_clone_table          0   32888
.data                    8   32888
.bss                     1   32896
.comment               184       0
Total                24058
```

</details>

### DIFF
--- a/examples/armv4t_multicore/emu.rs
+++ b/examples/armv4t_multicore/emu.rs
@@ -36,11 +36,10 @@ pub enum Event {
     WatchRead(u32),
 }
 
-#[derive(PartialEq)]
+#[derive(Debug, PartialEq)]
 pub enum ExecMode {
     Step,
     Continue,
-    Stop,
 }
 
 /// incredibly barebones armv4t-based emulator
@@ -49,6 +48,7 @@ pub struct Emu {
     pub(crate) cop: Cpu,
     pub(crate) mem: ExampleMem,
 
+    // If a CpuId is not in this, it is presumed "stopped".
     pub(crate) exec_mode: HashMap<CpuId, ExecMode>,
 
     pub(crate) watchpoints: Vec<u32>,
@@ -189,7 +189,7 @@ impl Emu {
         let mut evt = None;
 
         for id in [CpuId::Cpu, CpuId::Cop].iter().copied() {
-            if matches!(self.exec_mode.get(&id), Some(ExecMode::Stop)) {
+            if !self.exec_mode.contains_key(&id) {
                 continue;
             }
 
@@ -207,8 +207,7 @@ impl Emu {
         // The underlying armv4t_multicore emulator cycles all cores in lock-step.
         //
         // Inside `self.step()`, we iterate through all cores and only invoke
-        // `step_core` if that core's `ExecMode` is not `Stop`.
-
+        // `step_core` if that core has an `ExecMode`.
         let should_single_step = self.exec_mode.values().any(|mode| mode == &ExecMode::Step);
 
         match should_single_step {
--- a/examples/armv4t_multicore/gdb.rs
+++ b/examples/armv4t_multicore/gdb.rs
@@ -160,6 +160,10 @@ impl MultiThreadResume for Emu {
     }
 
     fn clear_resume_actions(&mut self) -> Result<(), Self::Error> {
+        // We're in all-stop mode (since `gdbstub` doesn't support non-stop yet), so
+        // the fact that we're processing commands from the GDB client means that all
+        // threads have already stopped, which is represented by the thread being
+        // absent from the `exec_mode` map.
         self.exec_mode.clear();
         Ok(())
     }
@@ -173,24 +177,26 @@ impl MultiThreadResume for Emu {
 
     fn set_resume_action_continue(
         &mut self,
-        tid: Tid,
+        tid: Option<Tid>,
         signal: Option<Signal>,
     ) -> Result<(), Self::Error> {
         if signal.is_some() {
             return Err("no support for continuing with signal");
         }
 
-        self.exec_mode
-            .insert(tid_to_cpuid(tid)?, ExecMode::Continue);
+        match tid {
+            None => {
+                for id in [CpuId::Cpu, CpuId::Cop] {
+                    self.exec_mode.entry(id).or_insert(ExecMode::Continue);
+                }
+            }
+            Some(tid) => {
+                self.exec_mode
+                    .insert(tid_to_cpuid(tid)?, ExecMode::Continue);
+            }
+        }
 
         Ok(())
-    }
-
-    #[inline(always)]
-    fn support_scheduler_locking(
-        &mut self,
-    ) -> Option<target::ext::base::multithread::MultiThreadSchedulerLockingOps<'_, Self>> {
-        Some(self)
     }
 }
 
@@ -298,15 +304,6 @@ impl target::ext::thread_extra_info::ThreadExtraInfo for Emu {
         let info = format!("CPU {:?}", cpu_id);
 
         Ok(copy_to_buf(info.as_bytes(), buf))
-    }
-}
-
-impl target::ext::base::multithread::MultiThreadSchedulerLocking for Emu {
-    fn set_resume_action_scheduler_lock(&mut self) -> Result<(), Self::Error> {
-        for id in [CpuId::Cpu, CpuId::Cop] {
-            self.exec_mode.entry(id).or_insert(ExecMode::Stop);
-        }
-        Ok(())
     }
 }
 
--- a/src/protocol/commands/_vCont.rs
+++ b/src/protocol/commands/_vCont.rs
@@ -25,7 +25,8 @@ impl<'a> ParseCommand<'a> for vCont<'a> {
         let body = buf.into_body();
         match body as &[u8] {
             b"?" => Some(vCont::Query),
-            _ => Some(vCont::Actions(Actions::new_from_buf(body))),
+            [b';', rest @ ..] => Some(vCont::Actions(Actions::new_from_buf(rest))),
+            _ => None
         }
     }
 }
@@ -52,7 +53,7 @@ impl<'a> Actions<'a> {
         Actions::FixedCont(tid)
     }
 
-    pub fn iter(&self) -> impl Iterator<Item = Option<VContAction<'a>>> + '_ {
+    pub fn iter(&self) -> impl DoubleEndedIterator<Item = Option<VContAction<'a>>> + '_ {
         match self {
             Actions::Buf(x) => EitherIter::A(x.iter()),
             Actions::FixedStep(x) => EitherIter::B(core::iter::once(Some(VContAction {
@@ -67,12 +68,16 @@ impl<'a> Actions<'a> {
     }
 }
 
+// This does not include the leading semicolon of the first action.
 #[derive(Debug)]
 pub struct ActionsBuf<'a>(&'a [u8]);
 
 impl<'a> ActionsBuf<'a> {
-    fn iter(&self) -> impl Iterator<Item = Option<VContAction<'a>>> + '_ {
-        self.0.split(|b| *b == b';').skip(1).map(|act| {
+    fn iter(&self) -> impl DoubleEndedIterator<Item = Option<VContAction<'a>>> + '_ {
+        // `ActionsBuf` doesn't include the leading semicolon in the first
+        // action, so we don't need to worry about the first element of the
+        // split being empty.
+        self.0.split(|b| *b == b';').map(|act| {
             let mut s = act.split(|b| *b == b':');
             let kind = s.next()?;
             let thread = match s.next() {
@@ -109,7 +114,7 @@ impl<'a> ActionsBuf<'a> {
                     //
                     // As a workaround for these weird GDB clients, `gdbstub`
                     // takes the pragmatic approach of treating this request as
-                    // though it the client requested _all_ threads to be
+                    // though the client requested _all_ threads to be
                     // resumed.
                     //
                     // If this turns out to be wrong... `gdbstub` can explore a
@@ -191,6 +196,20 @@ where
         match self {
             EitherIter::A(a) => a.next(),
             EitherIter::B(b) => b.next(),
+        }
+    }
+}
+
+impl<A, B, T> DoubleEndedIterator for EitherIter<A, B>
+where
+    A: DoubleEndedIterator<Item = T>,
+    B: DoubleEndedIterator<Item = T>,
+{
+    #[inline(always)]
+    fn next_back(&mut self) -> Option<T> {
+        match self {
+            EitherIter::A(a) => a.next_back(),
+            EitherIter::B(b) => b.next_back(),
         }
     }
 }
--- a/src/stub/core_impl/resume.rs
+++ b/src/stub/core_impl/resume.rs
@@ -89,6 +89,12 @@ impl<T: Target, C: Connection> GdbStubImpl<T, C> {
     ) -> Result<(), Error<T::Error, C::Error>> {
         use crate::protocol::commands::_vCont::VContKind;
 
+        // In the single-threaded scenario, we don't reverse the actions like we
+        // do in the multi-threaded: there are only two scenarios we concern
+        // ourselves with: 1 action, or 2 actions where the second is a
+        // continue action (sometimes GDB sends a packet of the form
+        // `vCont;s:foo;c`, even in single-threaded scenarios). We ignore the
+        // continue action, since there aren't any other threads to continue.
         let mut actions = actions.iter();
         let first_action = actions
             .next()
@@ -166,14 +172,16 @@ impl<T: Target, C: Connection> GdbStubImpl<T, C> {
     ) -> Result<(), Error<T::Error, C::Error>> {
         ops.clear_resume_actions().map_err(Error::TargetError)?;
 
-        // Track whether the packet contains a wildcard/default continue action
-        // (e.g., `c` or `c:-1`).
+        // NOTE: We iterate through these actions in reverse order, which corresponds to
+        // a right-to-left ordering of the actions specified in the vCont packet. This
+        // is intentionally the opposite of the left-to-right order specified by
+        // the vCont packet documentation.
         //
-        // Presence of this action implies "Scheduler Locking" is OFF.
-        // Absence implies "Scheduler Locking" is ON.
-        let mut has_wildcard_continue = false;
-
-        for action in actions.iter() {
+        // This is to simplify target implementations: each `set_resume_action_XXX`
+        // callback can overwrite the current state, instead of having to keep track of
+        // each thread specified by previous actions and making sure they don't get
+        // overwritten.
+        for action in actions.iter().rev() {
             use crate::protocol::commands::_vCont::VContKind;
 
             let action = action.ok_or(Error::PacketParse(
@@ -187,17 +195,15 @@ impl<T: Target, C: Connection> GdbStubImpl<T, C> {
                         _ => None,
                     };
 
-                    match action.thread.map(|thread| thread.tid) {
-                        // An action with no thread-id matches all threads
-                        None | Some(SpecificIdKind::All) => {
-                            // Target API contract specifies that the default
-                            // resume action for all threads is continue.
-                            has_wildcard_continue = true;
-                        }
-                        Some(SpecificIdKind::WithId(tid)) => ops
-                            .set_resume_action_continue(tid, signal)
-                            .map_err(Error::TargetError)?,
-                    }
+                    let tid = match action.thread.map(|thread| thread.tid) {
+                        // An action with no thread-id matches all threads, which is passed to
+                        // `set_resume_action_continue` as `None`.
+                        None | Some(SpecificIdKind::All) => None,
+                        Some(SpecificIdKind::WithId(tid)) => Some(tid),
+                    };
+
+                    ops.set_resume_action_continue(tid, signal)
+                        .map_err(Error::TargetError)?;
                 }
                 VContKind::Step | VContKind::StepWithSig(_)
                     if ops.support_single_step().is_some() =>
@@ -257,15 +263,6 @@ impl<T: Target, C: Connection> GdbStubImpl<T, C> {
                     return Err(Error::PacketUnexpected);
                 }
             }
-        }
-
-        if !has_wildcard_continue {
-            let Some(locking_ops) = ops.support_scheduler_locking() else {
-                return Err(Error::MissingMultiThreadSchedulerLocking);
-            };
-            locking_ops
-                .set_resume_action_scheduler_lock()
-                .map_err(Error::TargetError)?;
         }
 
         ops.resume().map_err(Error::TargetError)
--- a/src/target/ext/base/multithread.rs
+++ b/src/target/ext/base/multithread.rs
@@ -51,9 +51,9 @@ pub trait MultiThreadBase: Target {
     /// to indicate that memory starting at `start_addr + n` cannot be
     /// accessed.
     ///
-    /// Implemenations may also return an appropriate non-fatal error if the
+    /// Implementations may also return an appropriate non-fatal error if the
     /// requested address range could not be accessed (e.g: due to MMU
-    /// protection, unhanded page fault, etc...).
+    /// protection, unhandled page fault, etc...).
     ///
     /// Implementations must guarantee that the returned number is less than or
     /// equal `data.len()`.
@@ -67,7 +67,7 @@ pub trait MultiThreadBase: Target {
     /// Write bytes to the specified address range.
     ///
     /// If the requested address range could not be accessed (e.g: due to
-    /// MMU protection, unhanded page fault, etc...), an appropriate non-fatal
+    /// MMU protection, unhandled page fault, etc...), an appropriate non-fatal
     /// error should be returned.
     fn write_addrs(
         &mut self,
@@ -132,21 +132,21 @@ pub trait MultiThreadResume: Target {
     /// Upon returning from the `resume` method, the target being debugged
     /// should be configured to run according to whatever resume actions the
     /// GDB client had specified using any of the `set_resume_action_XXX`
-    /// methods.
+    /// methods. Any thread that wasn't explicitly specified in a
+    /// `set_resume_action_XXX` method should remain in the same state.
+    /// Currently, because `gdbstub` only supports all-stop mode, the only state
+    /// threads will be in while processing `resume` is stopped, so any threads
+    /// not specified in `set_resume_action_XXX` should remain stopped.
     ///
-    /// # Default Resume Behavior
+    /// If a thread is specified in multiple `set_resume_action_XXX` methods,
+    /// **the last call** takes precedence. A common situation when this happens
+    /// is if a `set_resume_action_XXX` method specifies all threads, and then
+    /// a following call to `set_resume_action_XXX` method specifies one
+    /// specific thread. The second call to the specific thread takes
+    /// precedence.
     ///
-    /// By default, any thread that wasn't explicitly resumed by a
-    /// `set_resume_action_XXX` method should be resumed as though it was
-    /// resumed with `set_resume_action_continue`.
-    ///
-    /// **However**, if [`support_scheduler_locking`] is implemented and
-    /// [`set_resume_action_scheduler_lock`] has been called for the current
-    /// resume cycle, this default changes: **unmentioned threads must remain
-    /// stopped.**
-    ///
-    /// [`support_scheduler_locking`]: Self::support_scheduler_locking
-    /// [`set_resume_action_scheduler_lock`]: MultiThreadSchedulerLocking::set_resume_action_scheduler_lock
+    /// Implementations may assume that at least one `set_resume_action_XXX` was
+    /// called between `clear_resume_actions` and `resume`.
     ///
     /// # Protocol Extensions
     ///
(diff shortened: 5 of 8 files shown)


src/target/ext/base/multithread.rs
```diff
@@ -202,14 +198,16 @@ pub trait MultiThreadResume: Target {
 
     /// Continue the specified thread.
     ///
+    /// If no thread is specified, continue all threads.
```
daniel5151, line 211: If we're making this change, lets also tweak the docs in `resume()` to describe the intended behavior when a `Tid`'s behavior is set by different `set_resume_action*` invocations (i.e: the "left-most vCont action takes priority" principle).

...and if we're doing that, changing the iteration order of the vCont actions iterator could be a good call as well, as IMO, explaining

> subsequent calls to `set_resume_action_*` may "override" earlier calls

is gonna be a lot easier for end users to implement vs. 

> if a `set_resume_action_*` corresponds to a thread that's already had it's behavior specified, you _must_ ignore it

Reversing iteration should be ezpz, simply by adding a `.rev()` into the code at https://github.com/daniel5151/gdbstub/blob/master/src/protocol/commands/_vCont.rs#L75

i.e:

```rs
// from
self.0.split(|b| *b == b';').skip(1).map(|act| { .. })
// to
self.0.split(|b| *b == b';').skip(1).rev().map(|act| { .. })
```

src/target/ext/base/multithread.rs
```diff
@@ -51,9 +51,9 @@ pub trait MultiThreadBase: Target {
     /// to indicate that memory starting at `start_addr + n` cannot be
     /// accessed.
     ///
-    /// Implemenations may also return an appropriate non-fatal error if the
+    /// Implementations may also return an appropriate non-fatal error if the
```
daniel5151, line 54: hehe, thank you 😅

examples/armv4t_multicore/gdb.rs
```diff
@@ -142,6 +142,10 @@ impl MultiThreadBase for Emu {
 
 impl MultiThreadResume for Emu {
     fn resume(&mut self) -> Result<(), Self::Error> {
+        if self.exec_mode.is_empty() {
+            return Err("received illegal empty vCont message");
+        }
```
daniel5151: I don't think we should nudge end users towards artificially inserting these sorts of assert in their code. I would imagine that most "real" implementations of `resume()` would look akin to `for action in action_list` (or some similar for-loop based construct), where the Target-level logic would be resilient in the case of not being given any actions.

That said, an empty vCont iterator is a cheap and easy to detect within `gdbstub` itself (possibly even at the packet parsing level itself? assuming it's not already a fatal parse error, it should be!), and document that from the perspective of the `resume()` method - one or more `set_resume_actions_*` calls would've occured prior to it being called

src/target/ext/base/multithread.rs
```diff
@@ -202,14 +198,16 @@ pub trait MultiThreadResume: Target {
...
+    ///
     /// See the [`resume`](Self::resume) docs for information on when this is
     /// called.
     ///
     /// The GDB client may also include a `signal` which should be passed to the
     /// target.
     fn set_resume_action_continue(
         &mut self,
-        tid: Tid,
+        tid: Option<Tid>,
```
daniel5151, line 220: This same transformation needs to apply to all other `set_resume_action_*` methods as well. i.e: AFAIK, there's nothing in the GDB RSP that states that you can't signal all threads to single step (ditto with the other resumption modes)

daniel5151 (commented):
Excellent first PR! A few minor comments, but at first blush - I don't see anything wrong with this approach.

And yeah, no worries about providing the big mess of logs. There are a lot of scenarios, and aside from trusting that you're validating the changes as you work on them, I'm sure I'll also spend some time poking around with the examples as we progress further along in `0.8`'s development

examples/armv4t_multicore/gdb.rs
jonathanzetier: Sounds good, yeah I didn't see anything in the documentation suggesting that `resume` was incapable of receiving no actions.

Looking at the `vCont` parsing code, it looks like there's currently no way for it to actually fail (https://github.com/daniel5151/gdbstub/blob/1bc505ff9ef71b0c08f15dc7e6b910b21ce885eb/src/protocol/commands/_vCont.rs#L24-L31); the actions don't get parsed until they're iterated through.  Knowingly passing around a `vCont` message that could be filled with nonsense makes me uncomfortable, but rewriting the parse logic seems a bit beyond the scope of this PR, so I'll check for an empty iterator and add that documentation.

My plan is to do this so that `clear_resume_actions` won't be called in the case of an empty iterator, to avoid stubs going into weird states, but this won't prevent weird states from malformed `vCont` data.

jonathanzetier:
Sounds good!  Typically with PRs after feedback, I amend my commit and force push, though I'm not sure how well GitHub handles this (most of my experience is with GitLab, which handles this...decently enough).  I know this is a controversial practice though; do you have any strong feelings about this?  I'm comfortable enough just pushing new commits if that's what you prefer.

daniel5151:
The PR is gonna get squash-merge'd, so you can feel free to iterate on it however you like!

examples/armv4t_multicore/gdb.rs
daniel5151: Yeah, the delayed parsing on the vCont packet is unfortunate... but past-me was hesitant to do a two-pass over it, hence the delayed failure mode. That said, I don't think it's _too_ bad, since any malformed packet means the entire debugging session is thrown out the window, so whether that happens before / after some calls to `set_resume_action_*` have been made isn't super consequential.

And rather than detecting an empty iterator - you should be able to add a cheap check in the current `_` arm to see if `body` is empty. The fact that isn't being done already is a bug, since we can and should be detecting (and bailing) on empty vCont packets

jonathanzetier:
After incorporating the changes to `vCont`, the binary size of the example_nostd went down a bit.  The `.text` section dropped a few hundred bytes, nothing to sneeze at (something like 2%), but that seemed to drop the size enough that there didn't need to be as much bytes in the relro padding section, which drops the size another 3000 bytes!

```
section               size    addr
.interp                 28     680
.note.ABI-tag           32     708
.note.gnu.build-id      36     740
.dynsym                384     776
.gnu.version            32    1160
.gnu.version_r          64    1192
.gnu.hash               28    1256
.dynstr                215    1284
.rela.dyn              432    1504
.rela.plt               24    1936
.rodata               1005    1968
.eh_frame_hdr          268    2976
.eh_frame             1328    3248
.text                15495    8672
.init                   27   24168
.fini                   13   24196
.plt                    32   24224
.fini_array              8   28352
.init_array              8   28360
.dynamic               432   28368
.got                   120   28800
.got.plt                32   28920
.relro_padding        3816   28952
.tm_clone_table          0   33048
.data                    8   33048
.bss                     1   33056
.comment               184       0
Total                24052
```

src/target/ext/base/multithread.rs
jonathanzetier: Alas, it wasn't quite as straightforward as the proposed change: the `skip` makes the `rev()` impossible, so I did a little bit of rejiggering the iterator to avoid the `skip`, so that it can be `rev()`-able.

I also felt like for iterating on the `_vCont` message itself, it seemed like it would make sense to still return a "normal", left-to-right iterator, since there might be other uses for the iterator.  Moving the `rev` to where we specifically need to swap the order made the intent a little clearer, at least to me.

For the single-threaded case, it took me a bit to understand what was going on here: 
https://github.com/daniel5151/gdbstub/blob/master/src/stub/core_impl/resume.rs#L93-L116

but from what I could figure out, we want either 1 action, or 2 actions as long as the second one is a `continue` action.  Doing this left-to-right with an iterator seems simpler to me, but if you'd like I can reverse it here too.  The reasoning behind why we allow 2 actions, as long as the second is a `continue` action, wasn't obvious to me (I still don't know), so I put a placeholder `XXX` in the comment explaining what's going on; do you remember the rationale for this decision? (does gdbserver do this/are there buggy gdb clients/is this an obscure part of the protocol?)  If so, let me know and I'll update the `XXX` placeholder to make it easier to look up than this PR comment :-)

src/protocol/commands/_vCont.rs
```diff
@@ -24,6 +24,9 @@ impl<'a> ParseCommand<'a> for vCont<'a> {
     fn from_packet(buf: PacketBuf<'a>) -> Option<Self> {
         let body = buf.into_body();
         match body as &[u8] {
+            // From https://sourceware.org/gdb/current/onlinedocs/gdb.html/Packets.html#vCont-packet:
+            // "Specifying no actions is an error."
+            b"" => None,
             b"?" => Some(vCont::Query),
             _ => Some(vCont::Actions(Actions::new_from_buf(body))),
```
daniel5151: I believe the following formulation will let you avoid the unnecessary empty-body handling in the iterator:

```suggestion
            [b';', rest @ ..] => Some(vCont::Actions(Actions::new_from_buf(rest))),
            _ => None,
```

src/stub/core_impl/resume.rs
```diff
@@ -89,6 +89,10 @@ impl<T: Target, C: Connection> GdbStubImpl<T, C> {
     ) -> Result<(), Error<T::Error, C::Error>> {
         use crate::protocol::commands::_vCont::VContKind;
 
+        // In the single-threaded scenario, we don't reverse the actions like we
+        // do in the multi-threaded: there are only two scenarios we concern
+        // ourselves with: 1 action, or 2 actions where the second is a
+        // continue action (which occurs when XXX?) that we ignore.
```
daniel5151: I _believe_ the behavior I observed in the past was that GDB would send over a packet along the lines of `vCont;s:foo,c`, even in single-threaded scenarios.

src/target/ext/base/multithread.rs
daniel5151: Responded in other comments :)

src/target/ext/base/multithread.rs
```diff
@@ -295,6 +296,8 @@ pub trait MultiThreadRangeStepping: Target + MultiThreadResume {
     /// and end (exclusive) addresses, or another stop condition is met
     /// (e.g: a breakpoint it hit).
     ///
+    /// If no thread is specified, step all threads, if supported.
```
daniel5151: I know I [just said](https://github.com/daniel5151/gdbstub/pull/197/changes/ff705650d824fc24ec1f3ffdf6b2463aeee64e77#r3196894945) that we should apply this transform to all functions... but now that I see this change in the context of range stepping, I wonder if this is actually the right call?

Sure, at a protocol level, this would be valid... but I wonder if it's actually possible (in practice) to configure GDB / LLDB to send over a range-step action as the fallback for all threads? Heck, same thought applies to single-step as a fallback action, and the reverse-step/continue actions as well.

Maybe for v0, we should leave all actions except for `set_resume_action_continue` as taking a specific `tid`, and adding some error-handling inside `gdbstub` to warn the user if the GDB client tries to use a non-continue fallback action (and, in turn - to open an issue on github reporting how they managed to get into that situation)?

At that point, assuming someone _does_ stumble across a valid debugging scenario with a non-continue default resume actions... it shouldn't be too hard to add new (backwards-compatible) `set_resume_action_{step,reange_step}_as_fallback` IDETs to allow hooking into that functionality.

The benefit is that in the common case where a client + target do _not_ support fallback actions aside from continue, the implementation of `set_resume_action_*` foo methods don't need to include error handling for `None` thread IDs.

jonathanzetier: Heh, that's why I left it off of the step resume actions in the first iteration, but figured it probably wasn't too much to ask of targets to complain about not being able to handle it.

The error message for `PacketUnexpected` (the original error returned in this case) seems to accomplish this, saying it's something we don't expect to happen, and a request to file an issue, so I'm planning to just revert these changes, unless you'd prefer a more specific error variant

daniel5151: `PacketUnexpected` is the "cop out" error message when past me was too lazy to add in a proper error message with more context, hah.

I'm fine using it here for expediency (and wouldn't block the PR on it), but if you want to add a proper error message with some more context, that might be better.

Error paths as a whole need more love, but that's a larger work stream in and of itself #112

src/protocol/commands/_vCont.rs
jonathanzetier: I thought about doing that at first too, but the protocol specifies that the leading semicolon is part of the action, so I figured keeping the semicolon in there is a more pure representation of the actions, and doesn't carry around the cognitive baggage of having to remember "this datatype has the full contents of every action *except the first one, which is missing the leading semicolon".

That being said, it's small enough and you could just as easily make the case that removing the leading semicolon cleans up the loop, and it's just shifting the cognitive baggage between the code versus datastructures, and it becomes a matter of preference.

Just wanted to offer an alternative viewpoint; I'll plan to implement the suggestion, unless you say otherwise!

daniel5151: Yeah, I think it's totally fine treating individual packet parser modules (i.e: `_vCont.rs`, etc...) as single "cognitive domains". And even though the iterator data structure does end up being used outside the module, external consumers don't need to care about how the parser works (i.e: whether it was front-loaded vs. deferred, etc...).

jonathanzetier:
In my last comment regarding the `cargo bloat` result, I seem to have accidentally copied the "before" size.  After the requested changes, the `cargo bloat` result is still significantly smaller (again due to the smaller `relro_padding` section). Here's what it is now (which is closer to what my last comment above was, if I had actually copied the correct size)

```
section               size    addr
.interp                 28     680
.note.ABI-tag           32     708
.note.gnu.build-id      36     740
.dynsym                384     776
.gnu.version            32    1160
.gnu.version_r          64    1192
.gnu.hash               28    1256
.dynstr                215    1284
.rela.dyn              432    1504
.rela.plt               24    1936
.rodata                973    1968
.eh_frame_hdr          260    2944
.eh_frame             1312    3208
.text                15008    8624
.init                   27   23632
.fini                   13   23660
.plt                    32   23680
.fini_array              8   27808
.init_array              8   27816
.dynamic               432   27824
.got                   120   28256
.got.plt                32   28376
.relro_padding         264   28408
.tm_clone_table          0   32504
.data                    8   32504
.bss                     1   32512
.comment               184       0
Total                19957
```

examples/armv4t_multicore/emu.rs
```diff
@@ -204,11 +204,18 @@ impl Emu {
     }
 
     pub fn run(&mut self, mut poll_incoming_data: impl FnMut() -> bool) -> RunEvent {
+        if self.exec_mode.is_empty() {
+            // This should never happen (gdbstub will always ensure at least one
+            // `set_resume_action_XXX` method is called), but in case it does, we explicitly
+            // log it and return the closest event that represents this.
+            eprintln!("Running while all threads are stopped; this should never happen!  Treating as 0 steps");
+            return RunEvent::Event(Event::DoneStep, CpuId::Cpu);
+        }
```
daniel5151: again, I _really_ think this entire block should just be omitted.

The docs and `gdbstub` implementation make it clear that the contract forbids calling this method without having called `set_resume_action_*`, so we shouldn't nudge end users to hedge against bugs in `gdbstub`.

If you want to leave this in purely in the example code, then swap it out with:


```suggestion
        assert!(!self.exec_mode.is_empty(), "gdbstub violated resume() contract (called prior to calling `set_resume_action_XXX` at least once");
```

src/protocol/commands/_vCont.rs
```diff
@@ -24,8 +24,13 @@ impl<'a> ParseCommand<'a> for vCont<'a> {
     fn from_packet(buf: PacketBuf<'a>) -> Option<Self> {
         let body = buf.into_body();
         match body as &[u8] {
+            // From https://sourceware.org/gdb/current/onlinedocs/gdb.html/Packets.html#vCont-packet:
+            // "Specifying no actions is an error."
+            b"" => None,
             b"?" => Some(vCont::Query),
-            _ => Some(vCont::Actions(Actions::new_from_buf(body))),
+            [b';', rest @ ..] => Some(vCont::Actions(Actions::new_from_buf(rest))),
+            // Anything that doesn't start with a semicolon is malformed
```
daniel5151: nix the comment (the `?` case is literally 2 lines up lol)

daniel5151 (commented):
2 small nits, but otherwise LGTM!

examples/armv4t_multicore/emu.rs
jonathanzetier: Whoops!  For some reason I thought you only meant the block in the `resume()` function, not this one.  I just removed the block entirely.

src/protocol/commands/_vCont.rs
jonathanzetier: Looking at it again, the empty case also doesn't need to exist, so I took that out too

daniel5151 (approved):
Nice! :shipit:

Issue #51, opened 2021-05-02
Expanding the `ConsoleOutput`-like interface


daniel5151:
At the moment, `gdbstub` only exposes `ConsoleOutput` in one place: as part of the `handle_monitor_cmd` interface. That said, this functionality of "print text from the target to the client terminal" isn't some special feature of the monitor command interface, and is also supported as a bog-standard [Stop Reply Packet](https://sourceware.org/gdb/current/onlinedocs/gdb/Stop-Reply-Packets.html#Stop-Reply-Packets) (namely, the `O XX...` packet).

Moreover, based on a very brief skim through the `gdb` client source code, it seems that the GDB client might be resilient enough to accept `O` packets at _any time_, not just as a response to `qRcmd` and as a stop reply packet! Note that this behavior is _not_ formalized by the spec, but if it works, then it opens the door to some very cool functionality...

With these bits of info in mind, there are a few ways `gdbstub` could expand expose `ConsoleOutput` in more places:

#### 1. Exposing `ConsoleOutput` as part of the current `resume()` interfaces.

This should be pretty straightforward, and could be implemented similar to the current monitor command `ConsoleOutput` (i.e: using a callback).

Note that this will require an API change to add the additional function parameter.

#### 2. Exposing a "global" handle to send `O XX...` packets.

At the time of writing, I don't have a concrete idea of how to implement this (especially for `no_std` targets), but I was thinking the API could look something like this:

```rust
let connection: TcpStream = wait_for_gdb_connection(9001);
let mut debugger = GdbStub::new(connection);

// get an instance of the global `ConsoleOutput` handle
let console_out_handle = debugger.console_output_handle();
let mut target = MyTarget::new(console_out_handle)?;
target.set_console_out_handle(console_out_handle); // or maybe the target supports late-binding

match debugger.run(&mut target) { ... }

// ..
// later, in the target implementation...
// ..

impl MultiThreadOps for MyTarget {
    fn read_registers(
        &mut self,
        regs: &mut gdbstub_arch::arm::reg::ArmCoreRegs,
        tid: Tid,
    ) -> TargetResult<(), Self> {
        outputln!(self.console_out_handle, "reading regs for tid {:?}", tid);
        // ...
    }
}
```

[...]

PR #198, opened 2026-05-09
Mega-batch of 0.8 breaking changes

### LINKED ISSUE #195: Migrate to type-safe stop reason APIs
Right now, users directly interface with the [BaseStopReason](https://docs.rs/gdbstub/latest/gdbstub/stub/enum.BaseStopReason.html) enum (and it's various type-aliases) to report stop reasons.

Unfortunately, this means `gdbstub` needs to include runtime checks to catch instances where a user attempts to report a stop reason they haven't claimed / implemented support for. e.g: passing a `HwBreak` stop reason, even through they haven't implemented hardware breakpoint support.

This isn't really a big deal in practice (it's not a particularly large foot-gun, and the error messages and docs make it clear how to remediate the issue), but given `gdbstub`'s focus on making illegal states unrepresentable... it'd be nice to fix up.

Here's one idea that I came up with:

Instead of:

```rust
// on Target
pub fn use_fork_stop_reason()

// on GdbStubStateMachineInner
pub fn report_stop(
    self,
    target: &mut T,
    reason: impl IntoStopReason<T>,
) -> Result<GdbStubStateMachine<'a, T, C>, GdbStubError<T::Error, C::Error>> { ... }
```

Consider this alternate formulation

```rust
// remove use_fork_stop_reason, add new trait a Target can impl
trait UseForkStopReason: Target {}

// on GdbStubStateMachineInner
pub fn report_stop(
    self,
    target: &mut T,
    reason: impl FnOnce(ReportStop<'_, '_, T, C>),
) -> Result<GdbStubStateMachine<'a, T, C>, GdbStubError<T::Error, C::Error>> { ... }

struct ReportStop<'a, 'b, T, C> {
    // stash a back-reference to state machine
    // (maybe better to just accept owned version + have impl FnOnce return it?)
    this: &'a mut GdbStubStateMachine<'b, T, C>
}

// where the type-safe magic happens
impl ReportStop<'_, '_, T, C> {
    fn done_step(self) { ... }
    fn exited(self, code: u8) { ... }
    // look! type-safe gating of methods, based on whether the target supports a particular trait
    fn hwbreak<Tid>(self, tid: Tid) where T: HwBreakpoints { ... }
    fn fork(self, curr_tid: Tid, new_tid: Tid) where T: UseForkStopReason { ... }
}
```

Of course, this is a _very_ rough sketch that I wanted to jot down while I had it in my mind... I did not check to see if it's actually feasible to implement in this current form, and whether the compiler would accept these sorts of trait bounds. If it works though... it could be really good!

Heck, it might even be possible to get rid of the enum representation of stop reasons entirely, and move to a function-per-stop-reason approach in the stub implementation itself. This has the potential to crunch the code down quite a bit (as well as making it a bit easier to read, I'd imagine?)

> Aside: as I wrote this out, I also realized that since IDETs are designed to be optionally runtime togglable (i.e: by returning `None` vs `Some(self)`), it may not be possible to remove runtime checks entirely. that said, in the common case where the Target unconditionally implements an IDET - those checks should be trivially optimized out in the handler functions (far better than in the case where an enum variant exists, but is unused).

Also, one more call-out: this approach may simplify the design of #185. It _might_ be possible to set things up such that the user-provided `name: &[u8]` data could be written directly into the response packet, without needing to maintain any long-lived reference to the data (and deal with the lifetime management issues inherent in that design).

daniel5151:
This is def the sort of thing that should've been multiple PRs, but I got a bit carried away... While I initially intended to split this up, the `Tid` and `StopReason` changes proved to be deeply intertwined at the state-machine layer, making a single cohesive PR the most practical approach.

**TL;DR of changes:**

* Added a `Tid` associated type to `Target`.
* Replaced `StopReason` enums with a fluent builder API.
* Reworked Ctrl-C interrupt handling to be asynchronous.
* Improved internal error types and state management.

Closes #195
Unblocks #51
Closes #185 (which I ended up implementing in this PR)

---

### Updated `Target` to include a thread-id associated type

A new `type Tid: IsValidTid` associated type has been added to the `Target` trait, and is now used throughout various IDETs, and internal/external `gdbstub` APIs.

Several things nudged me towards making this change:

* **Future-proofing for multi-process:** `gdbstub` has accumulated a couple non-`Base` IDETs that plumb thread-id values to the user's `Target` implementation (namely: `ThreadExtraInfo` and `Wasm`). These IDETs currently hard-code `crate::common::Tid` in their APIs. The jump to multi-process support will require updating their function signatures, and single-threaded targets that otherwise don't need to care about `Tid` values currently end up having `gdbstub`'s `SINGLE_THREAD_TID` constant leaked to them.
* **Compile-time safety:** While it *is* possible for particularly dynamic `Target` implementations to decide to swap between single-thread vs. multi-thread ops at runtime, in practice, a target picks what kind of target it is at compile time and sticks to it. Returning a non-`()` TID on a single-threaded target could confuse `gdbstub` internally. This change makes `Tid` handling *significantly* more consistent and type-safe across the board.
* **Internal simplification:** `gdbstub`'s implementation has been dramatically simplified and tangibly improved by having `T::Tid` everywhere. I've been able to refactor a surprisingly large amount of Tid conversion code. The new `from_fully_qualified_tid` adds strong type-level guard rails for validating that the GDB client is sending over thread-id values that the `Target` is able to operate on.
* **Reducing "generics soup":** Plumbing `Tid` via `Target` *significantly* reduces the amount of generics the `run_blocking` and `GdbStubStateMachine` family of APIs have to deal with. We no longer need an extra top-level `Tid: IsValidTid` bound on a bunch of those APIs.

### Removed `StopReason` enums, switched to function-based stop reason reporting

The `enum`-based API was workable so long as stop reasons remained fixed-length... but as we've implemented more of the GDB RSP, the stop reason API hasn't kept up with the increasingly dynamic data that folks want to report.

Two notable examples: reporting inline register values, and stop reasons with embedded strings. While we *were* able to squeeze in support for reporting inline register values via #189, things were getting a bit hacky, and would only get hackier as the GDB RSP added more metadata to stop reply packets (e.g: like how they added `core` to the `T` reply packet).

This new approach rethinks stop reason reporting from the ground up, moving away from the enum-based approach that was present since the earliest days of `gdbstub`, and swaps over to a more flexible, extensible, and *safer* fluent builder approach to reporting stop reasons.

A huge DX win here is **compile-time safety for IDETs**. The new builder methods (like `.swbreak()`) have trait bounds that will cause a compile-time failure if you try to use them without actually implementing the corresponding IDET on your target!

So, for example: in the context of the `run_blocking` infrastructure:

```rust
// Before
emu::RunEvent::Event(event) => {
    use gdbstub::target::ext::breakpoints::WatchKind;

    let stop_reason = match event {
        emu::Event::DoneStep => SingleThreadStopReason::DoneStep,
        emu::Event::Halted => SingleThreadStopReason::Terminated(Signal::SIGSTOP),
        // must carefully remember to only report SwBreak if corresponding
        // IDET is implemented
        emu::Event::Break => SingleThreadStopReason::SwBreak(()),
        ...
    };

    // no way to report inline registers :( would have to drop down
    // to the full-blown state machine API to do that
    Ok(run_blocking::Event::TargetStopped(stop_reason))
}

// After
emu::RunEvent::Event(event) => {
    Ok(simple_stub.report_stop(target, |report_stop| {
        use gdbstub::target::ext::breakpoints::WatchKind;

        match event {
            emu::Event::DoneStep => report_stop.done_step(),
            emu::Event::Halted => report_stop.terminated(Signal::SIGSTOP),
            // new APIs allow ergonomically chaining extra metadata
            // (e.g: via `add_reg`, or `add_core`) without having to move on
            // from the blocking event loop!
            // 
            // also, `swbreak` method has trait bounds that would
            // cause a compile-time fail if it was used without implementing
            // the corresponding IDET!
            emu::Event::Break => report_stop.swbreak(())?.done(),
            ...
        }
    }))
}

```

A major benefit under-the-hood is that there's no longer any point in time where a stop reason exists as a data structure in memory / needs to get "passed around" the codebase. This is a nice little binary size win! Albeit with one consequence...

### Rework Ctrl-C interrupt stop-reason reporting

The ctrl-c interrupt APIs now *require* the implementor to stash the fact the interrupt occurred somewhere on/around their `Target` implementation, and report the effects of the interrupt as a *downstream* stop reason the next time they enter the `Running` state.

This is a direct consequence of the shift to function-based stop reason reporting, as there is no longer an in-memory representation of a stop reason that can be stashed via the `deferred_stop_reason` infrastructure in the state machine.

While this means the "fast path" is gone, this is actually a solid architectural win. Forcing targets to stash the interrupt and handle it asynchronously in their own execution loops enforces a more robust architecture and prevents weird state desyncs between GDB and the target's actual execution state. In practice, the vast majority of target implementations already maintain some form of internal "pending interrupt" flag or event queue, so setting a simple boolean flag is a trivial adjustment.

### Internal Cleanup

* **Improved Error variants + display implementations:** The `InternalError` (and by extension `GdbStubError`) enum has been revamped. Catch-all errors like `TargetMismatch` have been replaced with highly specific variants like `UnexpectedIntegerSize`, `UnexpectedReg`, and `UnexpectedThreadId`. The `Display` implementation clearly delineates between "Client" errors and "Target/Implementation" errors.
* **Decoupled `ResponseWriterState`:** The `ResponseWriter` has been decoupled slightly by extracting its state into a `ResponseWriterState` struct. This allows the state to be passed around without needing a continuous mutable borrow over the underlying connection, which heavily facilitates the new `StopReasonReporter` builder pattern.
* In `ResponseWriterState`, collapsed two `bool`s into a bitflag (for a tiny bit of memory gainz)

Here is a draft for that section that matches the tone of the rest of the PR description. You could slot this in right before the "Internal Cleanup" section:

### Minor API terminology tweaks (`tid` -> `thread_id`)

While going through and updating the various `Target` extension traits to use the new `Self::Tid` associated type, I took the opportunity to clean up some of the parameter naming in the API docs.

Across the board, trait methods that previously accepted a `tid: Tid` parameter (like in `ThreadExtraInfo`, `Wasm`, `SingleRegisterAccess`, etc.) have been updated to use `thread_id: Self::Tid`.

This is a purely cosmetic change for implementors, but as we enter the world of multi-process support, being more intentional about `tid`, `pid`, and `thread-id` will be important to maintain grokkability of `gdbstub`'s large API surface.

---

### Quick Migration Checklist

What's surprising is that, even through there's a *lot* of churn in `gdbstub` itself, the actual end-user APIs that folks interact with aren't changing all that much from `0.7`. I'm running out of steam this weekend to write a comprehensive transition guide, but it shouldn't be too painful to move over!

At a high level, this is all that'll really be needed:

1. Adding `type Tid = ();` (or `type Tid = gdbstub::common::Tid;`) to existing `Target` impls.
2. Stop reason reporting will switch over to using functions instead of enums (a highly localized change).
3. Switch ctrl-c handling to set an internal flag + defer reporting the stop to the Running state


daniel5151:
@jonathanzetier I'm going to merge this big bundle of changes into the `dev/0.8` branch - apologies for the churn in whatever unpublished branches you might have 😅 

That said - I strongly believe that these changes (especially introduction of the `Target::Tid` associated type) will dramatically simplify enabling multi-process support, and makes it so that the compiler will play a helpful role in pointing out what parts of the codebase need to be "enlightened" to work in multi-process scenarios.

Please feel free to share any thoughts you have on these changes! I'm fairly certain that these changes are the right call, but if you spot anything that you think I might've blundered on, please let me know!

